PR #71, opened 2026-03-07
fix(import): preserve UTF-8 CJK text when decoding Kiwi strings

### LINKED ISSUE #69: Importing figma files will result in Chinese garbled characters
### Description

Importing figma files will result in Chinese garbled characters

Image

### Steps to reproduce

_No response_

### File

_No response_

### Platform

Web (app.openpencil.dev)

### Browser (web only)

_No response_

### Screenshots

_No response_

SergioChan:
Fixes #69

---

Chinese text corruption during .fig import traced to custom UTF-16→UTF-8 encoding logic in Kiwi `ByteBuffer.writeString`. This change switches string encoding to `TextEncoder`, while keeping Kiwi's null-terminated wire format.

### What changed
- Replaced manual UTF-8 byte assembly in `packages/core/src/kiwi/kiwi-schema/bb.ts` with `TextEncoder`-based encoding.
- Preserved the NUL-character guard (`\0`) to avoid invalid Kiwi strings.
- Added regression tests in `tests/engine/kiwi-bytebuffer-utf8.test.ts` for:
  - CJK-only string roundtrip
  - Mixed Unicode roundtrip (CJK + emoji + astral character)

### Validation
- ✅ `bun test tests/engine/kiwi-bytebuffer-utf8.test.ts`
- ⚠️ Did not run full `bun run check` in this cron run to keep scope/time minimal; this change is isolated to string codec behavior + dedicated regression tests.

**Checklist:**
- [x] Tests added or updated

### DIFF
--- a/packages/core/src/kiwi/kiwi-schema/bb.ts
+++ b/packages/core/src/kiwi/kiwi-schema/bb.ts
@@ -1,6 +1,7 @@
 let int32 = new Int32Array(1)
 let float32 = new Float32Array(int32.buffer)
 const textDecoder = new TextDecoder()
+const textEncoder = new TextEncoder()
 
 export class ByteBuffer {
   private _data: Uint8Array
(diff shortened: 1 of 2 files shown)


SergioChan:
Thanks for validating this on master — that’s very helpful. Agreed this likely isn’t a TextEncoder issue. I’ll pause this PR and investigate the .fig field decoding/rendering path from #69 instead so we can target the real root cause.

SergioChan:
Thanks for the review and validation. Closing this as superseded since the encoder path is not the root cause for #69. I will continue debugging the .fig import decode/render path and open a new narrowly scoped PR with a reproducible failing case.

PR #86, opened 2026-03-09
Fix auto-layout overflow, grid layout, AI chat UX

sld0Ant:
Fix layout overflow from AI-generated JSX, add grid layout support, and improve AI chat usability.

**Layout engine bugs (6 fixes):**
- Hidden children size preserved in Yoga tree
- FILL children use flexBasis(0) to share space from zero
- SPACE_EVENLY primary axis alignment
- Full layoutAlignSelf range (STRETCH, CENTER, END, BASELINE)
- Grid children stretch via setWidthStretch()/setHeightStretch()
- Absolute children configured with PositionType.Absolute

**Layout engine (base fixes):**
- Absolute positioning for elements with x/y inside auto-layout parents
- Text wrap (textAutoResize: 'HEIGHT') for fill-width text in auto-layout
- Yoga setMeasureFunc() for text nodes instead of static pre-measurement
- min/max constraints on flex containers and leaf nodes
- counterAxisAlignContent: SPACE_BETWEEN for wrapped layouts
- Cache text MeasureFunc results to avoid repeated buildParagraph calls

**Grid layout support:**
- JSX renderer: grid, columns, rows, columnGap, rowGap, gap props
- Grid child positioning: colStart, col, rowStart, row, colSpan, rowSpan
- Auto-height grids (no rows template)
- Flex children stretch to cell width in grid
- recomputeGridChild for correct flex-in-grid dimensions
- Grid-in-flex and flex-in-grid interaction fixes

**AI chat fixes:**
- Auto-enable flex when justify/items props are used (defence in depth)
- System prompt: require flex with justify/items, discourage export_image
- Fix ProviderSettings popover (tooltip/popover trigger conflict, z-index)
- Add configurable max output tokens (default 16384)
- Fix paste/copy/cut in chat input (event propagation)
- Strip TypeScript casts from AI-generated JSX

---

**Checklist:**
- [x] Tests added or updated

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -165,10 +165,10 @@ export const AI_PROVIDERS: AIProviderDef[] = [
     name: 'Google AI',
     keyPlaceholder: 'AIza…',
     keyURL: 'https://aistudio.google.com/apikey',
-    defaultModel: 'gemini-3.1-pro',
+    defaultModel: 'gemini-3.1-pro-preview',
     models: [
-      { id: 'gemini-3.1-pro', name: 'Gemini 3.1 Pro', tag: '1M context' },
-      { id: 'gemini-3-flash', name: 'Gemini 3 Flash', tag: 'Fast' }
+      { id: 'gemini-3.1-pro-preview', name: 'Gemini 3.1 Pro', tag: '1M context' },
+      { id: 'gemini-3-flash-preview', name: 'Gemini 3 Flash', tag: 'Fast' }
     ]
   },
   {


packages/core/src/render/render-jsx.ts
```diff
@@ -6,17 +6,27 @@ import { resolveToTree, type TreeNode } from './tree'
 
 import type { SceneGraph } from '../scene-graph'
 
+/**
+ * Strip TypeScript-only syntax that sucrase's JSX-only transform can't handle.
+ * AI models sometimes emit `as any`, `as const`, etc.
+ */
+function stripTypeAnnotations(jsx: string): string {
```
dannote: This regex is fragile — it'll break on string content containing "as any", miss complex casts like `as React.FC`, and silently mask real issues.

Better: just change the `transforms` array in `buildComponent` to `['typescript', 'jsx']` instead of `['jsx']`. Sucrase handles all TS syntax correctly. Drop `stripTypeAnnotations` entirely.

src/composables/use-chat.ts
```diff
@@ -116,15 +128,15 @@ const SYSTEM_PROMPT = dedent`
...
-  ⚠ Export ONE node at a time. Be extremely critical: check text clipping, spacing, alignment, contrast, overflow.
-  Fix any issues immediately, then re-export.
+  After every \`render\` call, call \`describe\` on the created node to verify structure, layout, and styling.
+  Be critical: check for missing props, wrong hierarchy, contrast issues.
+  Fix any issues immediately, then re-describe.
 
   # Reading designs
-  - \`export_image\`: renders to PNG for visual inspection
+  - \`describe\`: semantic description with role, style, layout, and design issues — preferred for verification
   - \`get_jsx\`: JSX representation (same format as render)
```
dannote: `export_image` is the only way the AI can visually verify colors, spacing, alignment, text contrast. `describe` gives structural info but can't catch visual bugs. The original workflow (render → export_image → verify → fix) is intentional. I'd revert this to encourage `export_image` for verification, not discourage it.

src/composables/use-chat.ts
```diff
@@ -165,6 +178,7 @@ watch(providerID, (id) => {
 watch(modelID, () => resetChat())
 watch(customModelID, () => resetChat())
 watch(customAPIType, () => resetChat())
+watch(maxOutputTokens, () => resetChat())
```
dannote: Resetting the entire chat when maxOutputTokens changes is bad UX — you lose your conversation just because you adjusted a setting. Remove this watch and apply the new value on the next message instead.

dannote (changes requested):
Lots of solid layout work — SPACE_EVENLY, extended layoutAlignSelf, text MeasureFunc, min/max constraints, absolute positioning in auto-layout, flexBasis(0) for FILL, grid-in-flex. Well-tested. The components.ts rest children fix and configureFlexContainer extraction are clean.

PR #94, opened 2026-03-10
MCP codegen pipeline: tools, prompt, and pi skill

sld0Ant:
3 new MCP tools for design-to-code workflow:

- `design_to_tokens` — extract Figma variables as CSS custom properties, Tailwind theme, or JSON. Resolves aliases, handles multi-mode (light/dark).
- `design_to_component_map` — analyze document structure: components (variants, props, instance counts), screens, sections.
- `get_codegen_prompt` — serve the 5-step codegen workflow prompt for AI agents (registered in MCP server only — core stays browser-safe).

Also includes:
- `packages/core/src/tools/prompts/codegen.md` — system prompt covering survey → decompose → extract tokens → generate code → verify

---

### AI token optimization (~61% reduction)

- System prompt: 19.5KB → 6.3KB — removed redundant rules, condensed layout/typography/spacing docs
- Tool tiers: `CORE_TOOLS` (34 tools, ~2.8K schema tokens) registered in AI chat vs `ALL_TOOLS` (98, ~6.5K) for MCP/CLI. Saves ~3,650 tokens per step.
- `<Icon>` JSX element: `<Icon name="lucide:heart" size={20} color="#FFF" />` fetches and renders inline. Eliminates search→fetch→insert×N→delete×N chain (~8 tool calls per design).
- `search_icons` default limit 20 → 5 (~300 fewer tokens per query)
- `get_page_tree` returns compact tree (id, type, name, w, h) instead of full recursive property dump
- `eval` returns `{ ok: true }` instead of `null`, saves a follow-up describe call
- `renderTree`/`renderJSX`/`renderTreeNode` now async to support inline Icon fetch

### AI debug log & prompt caching

- Tool execution log with before/after node snapshots — detects NO-OP mutations (tool succeeds but node properties unchanged)
- Duplicate tool call detection (mutating tools only — read-only `describe` calls expected)
- `buildDebugLog()` aggregate diagnostics: duplication rate, no-op count, payload sizes
- Anthropic prompt caching enabled via `cacheControl` in `ToolLoopAgent` — 94-98% cache hit rate, ~90% cost reduction on cached tokens

### AI chat UX

- Step limit raised from 20 → 50 (dating app session hit exactly 20 steps, with 10 consumed by icon searches)
- Step budget warnings injected into tool results when ~5 steps remain
- "Continue" button when AI reaches step limit — resumes with full context
- Block chat input during AI streaming to prevent message spam
- Provider/model/API key changes apply on next message without losing chat history
- Provider select dropdown z-index fix (no longer hidden behind settings popover)

### `calc` tool

Safe arithmetic evaluator for AI layout math — recursive descent parser (no eval/new Function). Supports `+ - * / % **`, parentheses, `Math.min/max/floor/ceil/round/abs/sqrt/pow`. Added to system prompt as mandatory before rendering into fixed-size containers.

### `set_layout` direction optional

When AI only changes spacing/padding, it no longer needs to re-state the current layout direction — eliminates ~50% of NO-OP mutations.

### `describe` tool improvements

- `depth` parameter (max 3) — `describeChild` recursively includes children, reducing 5-10 extra describe calls per render cycle
- Layout & visibility issue detection:
  - Text without `color` flagged as invisible
  - Shapes/icons without fill/stroke flagged as invisible
  - Dark text on dark background warning
  - `grow` inside HUG parent (no effect)
  - Children overflowing fixed-size parent
  - HUG container where all children use grow/stretch (collapses to 0)
  - Cross-axis overflow, text overflow, wrap without rowGap
  - Divider orientation vs parent layout direction
  - Spacing grid alignment, padding/gap consistency
  - Fill sizing without flex parent, absolute children in flex, nested flex without fill
  - Duplicate sibling names, stroke color/weight mismatch, circular frame without clip
  - Excessive nesting (3+ single-child wrappers), same fill as parent, image placeholder without content
  - Sibling height inconsistency, text wrapping detection, alignment issues
- Typography & text props: lineHeight, letterSpacing, textCase, textDecoration, maxLines
- Corner radius consistency (inner radius = outer − padding)
- Split into `describe.ts` + `describe-issues.ts` + `describe-layout-issues.ts`

### Icon rendering

- SVG shape primitives (`<circle>`, `<rect>`, `<ellipse>`, `<line>`, `<polygon>`, `<polyline>`) converted to path d-strings in iconify.ts. Fixes lucide:search (missing circle), lucide:user (missing head), lucide:mail (missing envelope).
- Icons placed at parent origin instead of viewport absolute coordinates after reparenting
- Batch insertion: `names` param (string array) inserts multiple icons in one call
- `search_icons` → `fetch_icons` prefetch pattern — search first, insert pre-fetched SVGs

### System prompt tuning

- Top-down rendering workflow: 3 renders + 3 describes, 40-element limit per render
- `depth=2` recommended for root frame describes
- Emoji guidance: canvas fonts lack emoji glyphs (□ tofu), use `<Icon>` for UI iconography
- `justify="evenly"` removed — not supported, explicit warning in prompt
- Flex requirement strengthened: every parent with fill children MUST have `flex="col"|"row"`
- Text always requires `color="#hex"`, shapes need bg or stroke
- Advanced tools available via `eval` — variables, boolean ops, path editing, analysis, codegen

### Layout engine

- `estimateTextSize()` fallback (~0.6 × fontSize per char) — prevents 100×100 text default from blowing up HUG containers
- Only call `computeAllLayouts` for mutating AI tools
- Fix text `w="fill"` stuck at 100px in `flex="col"` parent
- Fix fill text MeasureFunc constraint

### Lint cleanup

- 44 warnings + 2 errors → 15 warnings, 0 errors
- Fix nested ternary in `chat-debug.ts`
- Split `applyTextOverrides` (complexity 25→<20) into `applyTextStyleOverrides` + `applyTextAutoResize`
- Eliminate 27 `as any` casts: proper `Effect`, `StyleRun`, `VariableType`, `VariableValue`, `VectorNetwork` types
- Remove unnecessary casts in vector/create/modify/variables tools
- oxlint overrides for legitimate boundaries: `ai-adapter.ts` (valibot SDK), `render-jsx.ts` (JSX eval), `kiwi-schema` (vendored)

### PR review feedback

- Removed `docs/mcp-codegen-plan.md` (moved to [discussion #99](https://github.com/open-pencil/open-pencil/discussions/99))
- Removed `lol.fig` test fixture from repo root
- Moved `getCodegenPrompt` out of core into MCP server (core must be browser-safe)
- Removed `packages/mcp/skills/` (content saved to [issue #97](https://github.com/open-pencil/open-pencil/issues/97))


components.d.ts
```diff
@@ -11,7 +11,6 @@ export {}
 /* prettier-ignore */
```
dannote: Generated file — should this be in `.gitignore`?

packages/core/src/tools/describe-issues.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+const CONTAINER_TYPES = new Set(['FRAME', 'COMPONENT', 'INSTANCE'])
+const SHAPE_TYPES = new Set(['RECTANGLE', 'ELLIPSE', 'STAR', 'POLYGON', 'LINE'])
+const ICON_MAX_SIZE = 48
+const BUTTON_MAX_WIDTH = 200
+const BUTTON_MAX_HEIGHT = 50
+const BUTTON_MIN_HEIGHT = 28
+const BUTTON_MIN_RADIUS = 2
+
```
dannote: `looksLikeButton` and the associated constants (`BUTTON_MAX_WIDTH`, `BUTTON_MAX_HEIGHT`, `BUTTON_MIN_HEIGHT`, `BUTTON_MIN_RADIUS`, `CONTAINER_TYPES`) are duplicated between `describe.ts` and `describe-issues.ts`.

src/ai/tools.ts
```diff
@@ -3,16 +3,57 @@ import { tool } from 'ai'
 import * as v from 'valibot'
 
 import { makeFigmaFromStore } from '@/automation/figma-factory'
-import { ALL_TOOLS, computeAllLayouts, toolsToAI } from '@open-pencil/core'
+import { CORE_TOOLS, computeAllLayouts, toolsToAI } from '@open-pencil/core'
 
 import type { EditorStore } from '@/stores/editor'
-import type { SceneNode } from '@open-pencil/core'
+import type { SceneNode, StepBudget, ToolLogEntry } from '@open-pencil/core'
+
```
dannote: Module-level mutable state (`_toolLogEntries`, `_currentRunSteps`, `_stepUsages`) — the convention says "no module-level mutable state in components, use the editor store." This is not a component, but the same concern applies: these reset independently of the chat lifecycle. If two chats coexist or hot-reload happens, the log entries leak across sessions. Consider attaching these to the `Chat` instance or passing them through the adapter options.

src/ai/tools.ts
```diff
@@ -21,22 +62,31 @@ export function createAITools(store: EditorStore) {
         }
       },
       onAfterExecute: (def) => {
-        computeAllLayouts(store.graph, store.state.currentPageId)
-        store.requestRender()
-        if (def.mutates && beforeSnapshot) {
-          const before = beforeSnapshot
```
dannote: Good change — only running `computeAllLayouts` + `requestRender` for mutating tools. The old code did it on every read operation too, which was wasteful.

tests/engine/iconify.test.ts
```diff
@@ -0,0 +1,111 @@
+import { describe, expect, test } from 'bun:test'
```
dannote: All these tests hit the live Iconify API. They will flake in CI without network access or when the API is slow/down. Consider mocking the HTTP layer or at minimum skipping in offline CI.

PR #103, opened 2026-03-13
Fix Kiwi encodeColor crash on colors missing alpha field

sld0Ant:
Saving to .fig or copying to clipboard crashes with `Missing required field "a"` in `encodeColor` when any fill, stroke, effect, or gradient stop color lacks the `a` property at runtime.

Add `safeColor()` guard at the serialization boundary in `kiwi-serialize.ts` — defaults `a` to `1` when missing.

---

**Checklist:**
- [ ] Tests added or updated

### DIFF
--- a/packages/core/src/kiwi-serialize.ts
+++ b/packages/core/src/kiwi-serialize.ts
@@ -251,16 +251,20 @@ function exportTextData(node: SceneNode): NodeChange['textData'] {
   }
 }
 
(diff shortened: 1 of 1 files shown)


dannote:
Root cause fixed on master (7c63fc3): colors missing `a` entered via the FigmaAPI proxy setters (fills/strokes/effects). The proxy now normalizes all colors with `normalizeColor()` which defaults `a` to 1. The safeColor bandaid in kiwi-serialize is no longer needed.

PR #108, opened 2026-03-14
feat(mcp): add export_image_file tool and fix headless PNG rendering

mcdmags:
## Summary

Add `export_image_file` MCP tool and fix headless PNG rendering to match the editor.

## Problem

Three issues with headless image export:

1. **No file output** — `export_image` returns base64 inline. AI agents and CI pipelines need actual image files on disk.

2. **Broken flex layout** — `computeAllLayouts()` isn't called after `render` creates nodes via JSX, so Yoga never computes flex positioning. Everything collapses to top-left.

3. **Wrong text metrics** — Font loading via `fetch('/Inter-Regular.ttf')` fails silently in Node.js. Text measurement used a rough `0.6×` character-width heuristic that didn't account for font weight, kerning, or proportional widths. Result: text truncation (`$48,290` → `$48,29`), overlap (`%` wrapping), and misalignment.

### New tool: `export_image_file`
Saves PNG/JPG/WEBP to a file path. Parameters: `path` (required), `ids`, `format`, `scale`.

### Fixed export pipeline
```
ensureInterFont()                              # load Inter from ../fonts/ into cache
  → SkiaRenderer + loadFonts()                 # renderer picks up cached Inter
  → setTextMeasurer(renderer.measureTextNode)  # use ParagraphBuilder for measurement
  → computeAllLayouts()                        # Yoga uses accurate font metrics
  → renderNodesToImage()                       # render with same fonts that measured
```

The key insight: use `renderer.measureTextNode()` (ParagraphBuilder-based) as the layout engine's text measurer. Same SkiaRenderer instance, same `fontProvider`, same font shaping — measurement and rendering are guaranteed to agree. Zero hacks.

### `computeAllLayouts` in render tool
The `render` tool now calls `computeAllLayouts()` after creating nodes, so flex layout is resolved immediately — not deferred to export time.

## Changes

| File | Change |
|------|--------|
| `packages/mcp/src/server.ts` | Add `export_image_file` tool, `ensureInterFont()`, ParagraphBuilder text measurer |
| `packages/core/src/tools/create.ts` | Call `computeAllLayouts()` after `renderJSX` |

## Before / After

| | Before | After |
|---|--------|-------|
| KPI values | `$48,29` (truncated) | `$48,290` |
| Percentages | `%` wrapping to next line | `+12.5%` single line |
| Bold text | Overlapping characters | Clean rendering |
| Layout | Collapsed to top-left | Matches editor exactly |

## Test plan

- [x] `render` → `export_image_file` produces PNG on disk
- [x] PNG layout matches .fig editor rendering
- [x] Bold text (`$48,290`, `4m 32s`) renders correctly
- [x] Percentages (`+12.5%`, `-0.8%`) stay on single line
- [x] Activity feed text doesn't wrap incorrectly
- [x] `save_file` + `export_image_file` both succeed
- [x] `bun run check` — 0 errors (lint fix: removed unused import, top-level node: imports)

### DIFF
--- a/packages/core/src/tools/create.ts
+++ b/packages/core/src/tools/create.ts
@@ -62,14 +62,11 @@ export const render = defineTool({
   },
   execute: async (figma, args) => {
     const { renderJSX } = await import('../render/render-jsx.js')
-    const { computeAllLayouts } = await import('../layout.js')
     const result = await renderJSX(figma.graph, args.jsx, {
       parentId: args.parent_id ?? figma.currentPageId,
       x: args.x,
       y: args.y
     })
-    // Compute layout immediately so exported images reflect flex positioning
-    computeAllLayouts(figma.graph)
     return { id: result.id, name: result.name, type: result.type, children: result.childIds }
   }
 })
--- a/packages/core/src/render/renderer.ts
+++ b/packages/core/src/render/renderer.ts
@@ -2,6 +2,7 @@ import { parseColor, colorToFill } from '../color'
 import { TRANSPARENT } from '../constants'
 import { fetchIcons } from '../iconify'
 import { createIconFromPaths } from '../icon-render'
+import { computeAllLayouts } from '../layout'
 import { isTreeNode } from './tree'
 
 import type { SceneGraph, SceneNode, NodeType, LayoutMode, GridTrack, Stroke } from '../scene-graph'
@@ -91,6 +92,8 @@ export async function renderTree(
 
   if (options.x !== undefined) graph.updateNode(result.id, { x: options.x })
   if (options.y !== undefined) graph.updateNode(result.id, { y: options.y })
+
+  computeAllLayouts(graph)
 
   return {
     id: result.id,
--- a/packages/core/src/tools/index.ts
+++ b/packages/core/src/tools/index.ts
@@ -1,4 +1,5 @@
 export { ALL_TOOLS, CORE_TOOLS, EXTENDED_TOOLS } from './registry'
+export { exportImage } from './vector'
 export { defineTool, nodeToResult, nodeSummary } from './schema'
 export type { ToolDef, ParamDef, ParamType } from './schema'
 export { toolsToAI, buildDebugLog } from './ai-adapter'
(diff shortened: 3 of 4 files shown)


mcdmags:
## Visual comparison

### Before (headless export without fix)
Text truncation, overlap, and collapsed layout:
- `$48,290` renders as `$48,29` (last digit cut)
- `+12.5%` — the `%` wraps to next line
- `4m 32s` splits across two lines
- Bold text characters overlap

before

### After (with ParagraphBuilder text measurement)
PNG export now matches the .fig editor rendering exactly:
- All KPI values render in full on single lines
- Percentages stay inline
- Bold text renders cleanly
- Layout matches the editor

after

### Editor rendering (.fig opened in OpenPencil)
This is the ground truth — what the .fig looks like in the editor:

editor

---

> **Note to reviewer:** I'll replace the placeholder image URLs with actual screenshots shortly. The before/after comparison demonstrates the text measurement fix — the `ParagraphBuilder` approach guarantees measurement matches rendering since both use the same `SkiaRenderer` instance and `fontProvider`.

mcdmags:
image
@sld0Ant  ready for another review from updating with master but not sure how to fix this CI issue

packages/core/src/tools/create.ts
```diff
@@ -62,11 +62,14 @@ export const render = defineTool({
...
     const result = await renderJSX(figma.graph, args.jsx, {
       parentId: args.parent_id ?? figma.currentPageId,
       x: args.x,
       y: args.y
     })
+    // Compute layout immediately so exported images reflect flex positioning
+    computeAllLayouts(figma.graph)
     return { id: result.id, name: result.name, type: result.type, children: result.childIds }
   }
 })
```
dannote: `computeAllLayouts` should be called inside `renderJSX` itself (in `render-jsx.ts`), not in the tool. There are multiple callers — the `render` tool, `replace_node` in `structure.ts`, `automation/server.ts` — and they all need layout computed after rendering. Moving it into `renderJSX` means every caller gets it automatically instead of having to remember to add it.

Issue #114, opened 2026-03-15
Basic housekeeping functionalities? Login + persistant storage>


rosaboyle:
### Description

# Feature request: Supabase-backed auth and persistent storage for hosted deployments

## Problem

Hosted OpenPencil has no login and no cloud persistence. Designs live only in the browser session. Hard to reuse from other browser or another machine. 

## Proposal

Use **Supabase/Firebase** (Auth + Postgres + Storage) as the backend for both user authentication and document/asset persistence.

### Authentication

- Email/password, GitHub OAuth, Google OAuth via Supabase Auth
- Session persistence across browser restarts
- Sign-out clears local state (Including stuffs in localstorage)

### Document storage (Supabase Database)

`documents` table with: `id` (uuid), `owner_id` (FK to auth user), `title`, `data` (jsonb, full document tree), `thumbnail_path`, `created_at`, `updated_at`, `is_public`.

Core operations: save, list (gallery view with thumbnails), open by ID, delete. Auto-save with debounce as a follow-up.

### Asset storage (Supabase Storage)

Bucket `assets` with path `{user_id}/{document_id}/{filename}` for images, SVGs, fonts, and exports. Referenced by URL in document data instead of inlined data URIs.

### Sharing

`document_shares` table: `document_id`, `shared_with` (user), `permission` (view/edit). Public link sharing via `is_public` flag.

### Row Level Security

Owners have full CRUD. Shared users get scoped access. Public documents are read-visible to all. RLS enforced at the database level.

### MCP parity (#113)

MCP `new_document` / `save_file` / `open_document` operate on the same Supabase-backed store. Documents created via MCP appear in the browser gallery and vice versa.

## Configuration

```
NEXT_PUBLIC_SUPABASE_URL=https://your-project.supabase.co
NEXT_PUBLIC_SUPABASE_ANON_KEY=eyJ...
SUPABASE_SERVICE_ROLE_KEY=eyJ...
```

If no Supabase env vars are set, the app stays in local-only mode (no breaking change). When configured, a one-time import migrates existing localStorage designs into the cloud library.

## Reproduction of the gap

1. Create a design on Computer A.
2. Open OpenPencil on Computer B.
3. The design does not exist.

Expected: after logging in on both, the same designs are accessible.

## Related

- #113: this issue proposes a concrete Supabase-based implementation for that request.

sld0Ant:
[...]

Issue #113, opened 2026-03-15
Feature request: support shared persisted documents for hosted browser + MCP deployments


Gerry9000:
# OpenPencil feature request: hosted deployments need a shared document workflow between browser and MCP

## Summary

OpenPencil works well as a local-first product because the browser/app surface
and MCP/CLI workflows all operate on the same local document environment.

In a hosted deployment, that parity breaks if the browser app and MCP are
deployed as separate services without a shared document layer. We confirmed that
hosted MCP persistence works, but there is no corresponding browser-facing
document workflow that lets a document created via MCP be opened, edited,
shared, and then reopened via MCP as part of one continuous system.

This is a feature request for hosted/server-mode document parity with local
usage, without requiring the MCP service and browser app to run in the same pod
or process.

## What We Verified

- Hosted MCP health and protocol negotiation work
- Hosted MCP can create a document with `new_document`
- Hosted MCP can save a document with `save_file`
- A fresh hosted MCP session can reopen that same saved `.fig`

So the problem is not "hosted MCP cannot persist files."

The problem is that hosted MCP persistence is isolated from the browser app's
user-facing document workflow.

## Minimal Reproduction

1. Open the browser app in a hosted deployment.
2. In a separate MCP session, create and save a document.
3. Open a fresh MCP session and confirm that the same document can be reopened.
4. Try to find and open that same document from the browser app.

Expected:

- the browser app and MCP can operate on the same persisted document set

Actual:

- MCP persistence works
- the browser app has no corresponding documented/shared document workflow for
  those persisted MCP-created documents

## Why This Matters

The expected workflow parity is:

1. Create or edit a document via MCP
2. Open that same document in the browser app
3. Edit/share it in the browser
4. Reopen the updated document via MCP
5. Export/share the same document with teammates or external tools

In a local install, this is naturally possible because all surfaces operate on
the same local document world.

In a hosted deployment, there is currently no documented/shared document model
that makes this workflow possible without building custom glue outside
OpenPencil.

## Current Hosted Gap

What exists:

- Browser app
- Hosted MCP endpoint
- MCP-side file persistence

What is missing for hosted parity:

- shared document namespace between browser and MCP
- browser-visible list/open flow for MCP-created documents
- stable document identifiers instead of raw server file paths
- a documented persistence model for hosted use
- retrieval/handoff flows for other tools and automation

## Important Constraint

This request is **not** asking for "run the browser app and MCP in the same
pod" as the only solution.

[...]

PR #122, opened 2026-03-15
docs(deploy): add example Kubernetes deployment templates

Gerry9000:
## Summary

Example Kubernetes manifests for self-hosting OpenPencil, contributed from a production deployment. Intended as a reference starting point, not a prescriptive deployment method.

- **webapp-deployment.yaml** — Init containers fetch a release tarball and build with Bun; nginx serves the static PWA with TLS
- **mcp-deployment.yaml** — Init container installs `@open-pencil/mcp` from npm; nginx TLS sidecar with bearer auth injection
- **nginx-configmap.yaml** — Reverse proxy template for the MCP sidecar
- **README** — Quick start, env var reference, image options, security notes

All org-specific values are parameterized via `envsubst`. Default images use standard upstream (nginx:1.27-alpine, node:22-alpine); [DHI hardened images](https://docs.docker.com/dhi/) documented as an option. All containers run non-root with dropped capabilities and read-only root filesystems where possible.

Relates to #113, #114

### DIFF
--- a/deploy/k8s/README.md
+++ b/deploy/k8s/README.md
@@ -1,0 +1,90 @@
+# Kubernetes Deployment Templates
+
+Production-ready Kubernetes manifests for self-hosting OpenPencil.
+
+## Components
+
+- **webapp-deployment.yaml** — Web UI: fetches a release tarball, builds with Bun, serves via nginx with TLS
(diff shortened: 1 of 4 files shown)


dannote:
Thanks for putting this together! We haven't started thinking about deployment yet, so I'll come back to this once we get there. Keeping it open for reference.

dannote:
Closing for now — we're not at the deployment stage yet. Will revisit when we get there. Thanks!

PR #135, opened 2026-03-16
fix: kiwi serialization bugs causing broken auto-layout in Figma import

### LINKED ISSUE #134: Auto-layout .fig exports broken when imported into Figma
## Problem

.fig files exported by OpenPencil with auto-layout containers render incorrectly when imported into Figma. This effectively makes OpenPencil unusable for creating designs that will be opened in Figma — which is the primary use case.

## Symptoms

When a .fig file containing auto-layout frames is opened in Figma:

- **Text nodes overlap** — all text collapses to the same position instead of being spaced by the parent's auto-layout
- **Auto-layout children stack at wrong positions** — children retain their absolute coordinates instead of being positioned by the layout engine
- **Text height computes as 0** — text nodes show dimensions like "343 Hug × 0 Hug" because Figma can't determine line height
- **Content overflows frame boundaries** — frames that should clip content don't

## Root Causes

Four bugs in `packages/core/src/kiwi-serialize.ts`:

1. **Auto-layout children get absolute transforms instead of (0,0)** — Figma expects auto-layout children to have identity transforms because it computes their positions at render time. OpenPencil was writing the in-memory `(node.x, node.y)` coordinates, causing Figma to double-offset children.

2. **Missing `bordersTakeSpace` field** — The `bordersTakeSpace` kiwi field (schema field 294) was never written during export, so Figma couldn't determine whether strokes affect layout sizing.

3. **Missing `lineHeight` on text nodes** — Text nodes without an explicit `lineHeight` got no lineHeight in the kiwi output. Without lineHeight (and without baselines/glyphs data), Figma computes 0 height for text, causing auto-layout children to collapse to the same position.

4. **Font family mismatch in `derivedTextData`** — The `fontMetaData` key in `derivedTextData` used the raw font family name (e.g., "DM Sans 9pt") while `fontName.family` used the normalized name ("DM Sans"), causing font lookup failures.

## Reproduction

1. Create a vertical auto-layout frame with 3+ text children using the MCP `render` tool
2. Save as .fig
3. Open in Figma
4. Observe: all text nodes overlap at position (0,0) with 0 height

## Environment

- OpenPencil: current master (commit f6f0765)
- Figma: Desktop app (latest)
- OS: Windows 11 / WSL2

rubinsh:
Fixes #134

## Summary

Auto-layout .fig files exported by OpenPencil were broken when imported into Figma — text overlapped, children stacked at wrong positions, and text heights computed as 0. This PR fixes 4 bugs in the kiwi serializer that caused these issues.

## Changes

Changes in `packages/core/src/kiwi-serialize.ts` and `packages/core/src/fonts.ts`:

- **Zero transforms for auto-layout children**: Children of auto-layout frames now get `(0, 0)` transforms instead of their absolute coordinates. Figma computes child positions at render time, so non-zero transforms caused double-offsetting. Absolute-positioned children and non-auto-layout children are unaffected.

- **Write `bordersTakeSpace`**: The `bordersTakeSpace` kiwi field (schema field 294) is now written from `node.strokesIncludedInLayout`. Previously missing entirely.

- **Always write `lineHeight` for text nodes**: Text nodes now always include explicit `lineHeight`, defaulting to `Math.ceil(fontSize * 1.2)` when not set. Without this, Figma computed 0 height for text (no baselines/glyphs data to fall back on).

- **Normalize font family in `derivedTextData`**: `fontMetaData` key now uses the normalized font family name (matching `fontName.family`) instead of the raw name with optical size suffixes. Also extended `normalizeFontFamily()` to strip optical size suffixes (e.g., "DM Sans 9pt" → "DM Sans").

- **Extract `computeExportTransform()` helper**: Refactored transform computation into a separate function for lint compliance (complexity limit).

## Tests

20 new unit tests in `tests/engine/kiwi-serialize-fixes.test.ts`:
- 4 tests for transform zeroing (auto-layout, absolute, non-auto-layout, horizontal)
- 4 tests for frameMaskDisabled behavior
- 3 tests for bordersTakeSpace
- 4 tests for lineHeight (explicit, default, odd fontSize, roundtrip)
- 4 tests for font normalization
- 1 integration test (full StatCard component)

## Verification

Exported a 24-component design system .fig file through the patched serializer and imported into Figma Desktop. All auto-layout frames render correctly with proper spacing, text sizing, and child positioning.

### DIFF
--- a/packages/core/src/fonts.ts
+++ b/packages/core/src/fonts.ts
@@ -65,7 +65,7 @@ const googleFontsCache = new Map<string, Record<string, string>>()
 const googleFontsFailed = new Set<string>()
 
 export function normalizeFontFamily(family: string): string {
-  return family.replace(/\s+Variable$/i, '')
+  return family.replace(/\s+(Variable|\d+(?:pt|px|em))$/i, '')
 }
 
 async function fetchGoogleFontFiles(family: string): Promise<Record<string, string> | null> {
--- a/packages/core/src/kiwi-serialize.ts
+++ b/packages/core/src/kiwi-serialize.ts
@@ -2,13 +2,13 @@ export const FIG_KIWI_VERSION = 106
 
 import { deflateSync, inflateSync } from 'fflate'
 
-import { weightToStyle, getLoadedFontData } from './fonts'
+import { weightToStyle, getLoadedFontData, normalizeFontFamily } from './fonts'
 import { encodeVectorNetworkBlob } from './vector'
 import { stringToGuid, VARIABLE_BINDING_FIELDS } from './kiwi/kiwi-convert'
 
 import type { NodeChange, Paint, VariableConsumptionEntry } from './kiwi/codec'
 import type { SceneGraph, SceneNode, CharacterStyleOverride } from './scene-graph'
-import type { Color, GUID } from './types'
+import type { Color, GUID, Matrix } from './types'
 
 const fontDigestCache = new Map<string, Uint8Array>()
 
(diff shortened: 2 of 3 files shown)


packages/core/src/fonts.ts
```diff
@@ -65,7 +65,7 @@ const googleFontsCache = new Map<string, Record<string, string>>()
 const googleFontsFailed = new Set<string>()
 
 export function normalizeFontFamily(family: string): string {
-  return family.replace(/\s+Variable$/i, '')
+  return family.replace(/\s+(Variable|\d+(?:pt|px|em))$/i, '')
```
rubinsh: this fix is a part of #132 - I had to include it here for my tests - but it shouldn't cause merge conflicts IMO

PR #140, opened 2026-03-17
JSX editor in Code panel

### LINKED ISSUE #130: Paste JSX / JSX Editor
### Description

Paste valid JSX code / Allow the JSX panel to be editable

### Use case

When unable/unwilling to connect the AI as an integration, it would be useful if you could edit the JSX directly and/or paste JSX in. This is useful where power users may just want to code out the design in JSX or Tailwind, or want their LLM to generate pastable samples.

### Figma reference

_No response_

sld0Ant:
Editable JSX in the Code panel — write, paste, or edit OpenPencil JSX to create and update design nodes directly. No AI integration needed.

**Editor mode:**
- "Write JSX" button in empty state, pencil icon in read-only view (OpenPencil format only)
- Syntax-highlighted textarea with line numbers and placeholder
- Apply (+ Cmd/Ctrl+Enter), Cancel (+ Esc), Tab inserts 2 spaces
- Inline error display on parse/render failure
- Editing existing selection replaces original nodes at the same position
- "Copy JSX Reference" button (📖) copies full tag/prop/example reference to clipboard

**Core changes:**
- `renderJSX` now returns `RenderResult[]` — supports multi-root JSX via fragment wrapping
- `buildComponent` adds `Component` and `Instance` tag aliases
- `renderJSXWithUndo` in editor store with full undo/redo
- Guard keyboard shortcuts when focus is in input/textarea (`useMagicKeys` fired delete/tool shortcuts while typing)

Fixes #130

---

**Checklist:**
- [x] Tests added or updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features
 
+- JSX editor in Code panel — write or paste OpenPencil JSX to create design nodes, with syntax highlighting, undo support, and error display
+- "Copy JSX Reference" button in JSX editor — copies the full prop/tag reference to clipboard for LLM prompts
 - Lock and visibility toggle buttons in layers panel (hover to reveal, always shown when active)
 - Figma-style selection scope — double-click to enter groups/frames/components, Escape to exit
 - Nested container navigation — each double-click goes one level deeper
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -191,7 +191,8 @@ export {
   type RenderResult,
   sceneNodeToJSX,
   selectionToJSX,
-  type JSXFormat
+  type JSXFormat,
+  JSX_REFERENCE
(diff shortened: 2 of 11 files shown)


gcoulby (changes requested):
There is a visual bug when the length of code added exceeds the height of the browser 

Image

I was able to paste design in. However, when pasted in large blocks the UI bugs out on Edge browser (Version 146.0.3856.78 (Official build) (64-bit). The panel breaks and you have to use Ctrl+Enter to submit it. Once done the design shows up, but the panel remains broken. 

Adding the following changes in the inspector fixed the issue, so this seems like an easy fix

Image

dannote:
Took the best parts — JSX reference as markdown, multi-root JSX support, Component/Instance aliases, copy reference button. Thanks! 💛

PR #143, opened 2026-03-17
Guard keyboard shortcuts during canvas text editing

### LINKED ISSUE #142: (minor bug) Deleting text inside Text Labels results in deleting a whole element
### **what is wrong?**
Anytime you add a text label into design file and try to write anything in text editing mode, when pressing `Backspace` or `Delete`, I was expecting to delete a specific character(-s) I've selected, not the whole label element. Maybe system doesn't have or recognize states between editing and selecting.

Happens on both Windows 11 (tauri) and online versions. I don't own Linux or macOS, so couldn't test on those machines.

(Also noticed that something weird happens as well when trying to add exclamation mark or at sign with `LShift` + `1` and `LShift` + `2` accordingly, it zooms out weirdly instead, probably expecting _Zoom to Fit_ and _Zoom to Selection_ to work) .

### Steps to reproduce

1. Open / Create a new .fig file inside OpenPencil application or online
2. Add a new Text Label
3. Edit any text inside there
4. While editing, deleting character(-s) by pressing `Backspace` or `Delete` keys leads into deleting a whole element

### File

_No response_

### Platform

Windows (Tauri)

### Browser (web only)

_No response_

### Screenshots

https://github.com/user-attachments/assets/186c94b0-4928-4ade-a0d2-26d59cec4c8a

just FYI: characters are not explicitly required to be selected, it's enough to just press "deleting" keybinds for this to take effect

sld0Ant:
Fixes #142

Backspace/Delete and tool shortcuts (V, R, F, T, etc.) fired during canvas text editing because the guard only checked for DOM input/textarea focus (`document.activeElement`), not the canvas TextEditor state (`editingTextId`).

Two places guarded:
- `onEventFired` — prevents tool switching (typing "v" switched to Move tool)
- `plain()` helper — prevents Backspace/Delete from calling `deleteSelected()`

---

**Checklist:**
- [ ] Tests added or updated

### DIFF
--- a/src/composables/use-keyboard.ts
+++ b/src/composables/use-keyboard.ts
@@ -95,6 +95,7 @@ export function useKeyboard() {
     onEventFired(e) {
       if (e.type !== 'keydown') return
       if (isEditing(e)) return
+      if (store.state.editingTextId) return
 
       if (!e.metaKey && !e.ctrlKey && !e.altKey) {
         const tool = TOOL_SHORTCUTS[e.key.toLowerCase()]
@@ -194,7 +195,8 @@ export function useKeyboard() {
         !keys['meta'].value &&
         !keys['control'].value &&
         !keys['shift'].value &&
-        !keys['alt'].value
+        !keys['alt'].value &&
+        !store.state.editingTextId
     )
   }
 


dannote:
Thanks for this — I merged it into master while reconciling the refactor, so closing this PR as already integrated. Really appreciate the contribution 💛

PR #161, opened 2026-03-27
Add GLM-5.1 model id

almontasser:
Adds the newly released models:
- GLM-5.1 (As of now the model only works on the Anthropic-compatible endpoint, so the model will not work right now)
- MiniMax-M2.7
- MiniMax-M2.7-highspeed
---

**Checklist:**
- [x] Tests added or updated


almontasser:
@dannote For some odd reason the Anthropic-compatible endpoint is not working for me, the old OpenAI-compatible endpoint works great, they've also added the new GLM-5.1 model

PR #166, opened 2026-03-31
fix: add missing StackWrap enum and fields to kiwi schema

### LINKED ISSUE #165: layoutWrap and counterAxisSpacing silently dropped on .fig save
## Summary

When a frame has `layoutWrap: 'WRAP'` (flex-wrap) and/or `counterAxisSpacing` (row-gap), these properties are lost after saving to `.fig` and reloading. The serialize/deserialize code correctly handles these values, but the kiwi binary schema does not define the fields — so the kiwi encoder silently drops them.

## Reproduction

1. Create a frame with `flex="row" wrap gap={8} rowGap={8}`
2. Add enough children to trigger wrapping (e.g. 12 chip elements in a 274px container)
3. Verify chips wrap to multiple lines
4. Save as `.fig`
5. Reload the `.fig` file
6. **Result:** Chips are in a single horizontal row — `wrap` is gone

## Root Cause

The serialize path (`kiwi-serialize.ts`) correctly writes `stackWrap` and `stackCounterSpacing` to the `NodeChange` object. The deserialize path (`kiwi-convert.ts`) correctly reads them back.

However, `kiwi/schema.ts` does not define these fields in the `NodeChange` message. The kiwi binary encoder only serializes fields that exist in the schema definition. Since `stackWrap` and `stackCounterSpacing` are not in the schema, they are silently ignored during encoding.

Round-trip:
```
SceneNode.layoutWrap = 'WRAP'
  → nc.stackWrap = 'WRAP'        ← serialize sets it
    → kiwi encode → (dropped)    ← schema doesn't know this field
      → kiwi decode → nc.stackWrap = undefined
        → layoutWrap = 'NO_WRAP'  ← deserialize defaults to NO_WRAP
```

## Fix

Add `StackWrap` enum and both missing fields to the kiwi schema in `packages/core/src/kiwi/schema.ts`:

```diff
+enum StackWrap {
+  NO_WRAP = 0;
+  WRAP = 1;
+}

 message NodeChange {
   // ... existing fields ...
   GridTrackSize[] gridColumnSizes = 474;
   GridTrackSize[] gridRowSizes = 475;
+  StackWrap stackWrap = 476;
+  float stackCounterSpacing = 477;
 }
```

Additionally, the `StackAlign` enum is missing the `STRETCH` value needed for cross-axis stretch alignment in auto-layout.

## Impact

Any `.fig` file saved with wrap-enabled auto-layout frames will lose:
- `layoutWrap: 'WRAP'` → reverts to `NO_WRAP` (single row/column)
- `counterAxisSpacing` → reverts to `0` (no row-gap in wrap layouts)

This affects both the editor save/load and CLI export/import paths.

## Before / After

Image
<p> -> </p>
Image

greekr4:
## Summary

- Add `StackWrap` enum (`NO_WRAP`, `WRAP`) to kiwi schema
- Add `stackWrap` (field 476) and `stackCounterSpacing` (field 477) to `NodeChange` message
- Add `STRETCH` value (4) to `StackAlign` enum

## Problem

`layoutWrap` (flex-wrap) and `counterAxisSpacing` (row-gap) are silently dropped when saving to `.fig` format. The serialize/deserialize code in `kiwi-serialize.ts` and `kiwi-convert.ts` already handles these fields correctly, but the kiwi binary schema didn't define them — so the encoder ignores them during encoding.

Any `.fig` file saved with wrap-enabled auto-layout frames loses wrap and row-gap on reload.

## Test plan

- [ ] Create a frame with `layoutWrap: 'WRAP'` and `counterAxisSpacing > 0`
- [ ] Save as `.fig`, reload, verify wrap and spacing are preserved
- [ ] Run `bun run test:unit` — existing tests should pass

Fixes #165

### DIFF
--- a/packages/core/src/kiwi/schema.ts
+++ b/packages/core/src/kiwi/schema.ts
@@ -314,6 +314,7 @@ enum StackAlign {
   CENTER = 1;
   MAX = 2;
   BASELINE = 3;
+  STRETCH = 4;
 }
 
 enum StackCounterAlign {
@@ -341,6 +342,11 @@ enum StackSize {
 enum StackPositioning {
   AUTO = 0;
   ABSOLUTE = 1;
+}
+
+enum StackWrap {
+  NO_WRAP = 0;
+  WRAP = 1;
 }
 
 enum ConnectionType {
@@ -1506,6 +1512,8 @@ message NodeChange {
   float gridColumnGap = 438;
   GridTrackSize[] gridColumnSizes = 474;
   GridTrackSize[] gridRowSizes = 475;
+  StackWrap stackWrap = 476;
+  float stackCounterSpacing = 477;
 }
 
 message VideoPlayback {


greekr4:
## Before / After

### ✅ After fix — `layoutWrap: 'WRAP'` preserved
SCR-20260331-kcre

Chips correctly wrap to multiple lines with `counterAxisSpacing` (row-gap) intact.

### ❌ Before fix — wrap lost on reload
KakaoTalk_Photo_2026-03-31-11-08-58

After saving and reloading the `.fig` file, all chips collapse into a single horizontal row — `layoutWrap` reverts to `NO_WRAP`.

PR #191, opened 2026-04-10
Fix tab order and keyboard handling in inspector panel

sld0Ant:
Closes #180

- ScrubInput receives focus via Tab (`tabindex=0`), auto-enters edit mode on focus
- All inspector buttons get explicit `tabindex=0` matching DOM/visual order
- ScrubInput values apply live while typing (`@input` → `update:modelValue`)
- Keyboard shortcuts (Backspace, tool keys, Enter) blocked while ScrubInput is focused

### DIFF
--- a/src/components/DesignPanel.vue
+++ b/src/components/DesignPanel.vue
@@ -71,14 +71,14 @@ const { panels } = useI18n()
       v-if="node.type === 'INSTANCE'"
       class="flex flex-col gap-1 border-b border-border px-3 py-2"
     >
-      <button tabindex="0"
+      <button
         data-test-id="design-go-to-component"
         class="rounded bg-component/10 px-2 py-1 text-left text-[11px] text-component hover:bg-component/20"
         @click="goToMainComponent.run()"
       >
         {{ panels.goToMainComponent }}
       </button>
-      <button tabindex="0"
+      <button
         data-test-id="design-detach-instance"
         class="rounded px-2 py-1 text-left text-[11px] text-muted hover:bg-hover"
         @click="detachInstance.run()"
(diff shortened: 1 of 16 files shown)


src/composables/use-keyboard.ts
```diff
@@ -193,6 +198,12 @@ export function useKeyboard() {
     )
   }
 
+  function isScrubInputFocused() {
+    const el = document.activeElement
+    if (!el) return false
+    return el.closest('[data-test-id="scrub-input"]') != null
+  }
+
```
dannote: Querying the DOM by test attribute to determine app state is not really the Vue way. `ScrubInputRoot` already has reactive `editing` state — it just doesn't surface it to the parent.

The editor store should track this the same way it tracks `editingTextId`. Then `plain()` just checks `store.state.scrubInputFocused` instead of `document.activeElement` + `.closest()`.

Same applies to the `.closest()` addition in `isEditing()` above.

src/components/DesignPanel.vue
```diff
@@ -71,14 +71,14 @@ const { panels } = useI18n()
       v-if="node.type === 'INSTANCE'"
       class="flex flex-col gap-1 border-b border-border px-3 py-2"
     >
-      <button
+      <button tabindex="0"
```
dannote: `<button>` is natively focusable — it already has implicit `tabindex="0"`. Adding it explicitly is a no-op. Same for all the other `tabindex="0"` additions on buttons throughout this PR.

The actual tab order fix is the ScrubInput wrapper `<div>` change (making a non-interactive element focusable) — that part is correct. But these button changes don't do anything.

dannote (changes requested):
Two issues with the current approach — details in inline comments.

sld0Ant:
Hi! Thanks for the review, all points are fixed.

dannote:
Merged, thanks for the contribution 💛

PR #195, opened 2026-04-13
fix(hit): fix hit area when parent rotation

yoohoo360:
Fixes #

### Problem

The selection box is not displayed correctly.

image

Your hover state is also not displaying correctly here.

### Fixed

image

image

### Changes

- Improved hit-testing behavior for common scenarios when the parent frame is rotated, and fixed rendering issues with the selection box.
- Added a test demo, including an example for saving demo graphics.
- Introduced matrix-based transformations to ensure accurate local and global positioning, avoiding precision errors.
- Added hot zones and cursor styles for rotation and resize when an element is selected.

---

**Checklist:**
- [x] Tests added or updated


packages/vue/src/shared/input/resize.ts
```diff
@@ -79,25 +86,39 @@ export function applyResize(
...
   const newW = Math.round(Math.max(1, width))
   const newH = Math.round(Math.max(1, height))
+  const newX = Math.round(x)
+  const newY = Math.round(y)
+
+  if (DEBUG_RESIZE) {
+    const dw = newW - Math.round(origW)
+    const dh = newH - Math.round(origH)
+    const dxyX = newX - Math.round(origX)
+    const dxyY = newY - Math.round(origY)
```
dannote: Debug code left in — `DEBUG_RESIZE = true` with `console.debug` must be removed.

packages/vue/src/shared/input/geometry.ts
```diff
@@ -77,6 +83,9 @@ export function unrotate(
   }
 }
 
+/**
```
dannote: `@deprecated` on functions with no external consumers is wrong — just delete the old functions and update call sites.

packages/core/src/scene-graph/hit-test.ts
```diff
@@ -1,6 +1,7 @@
-import { degToRad } from '../geometry'
+import { getWorldMatrix } from '../canvas/coordinate'
 
 import type { SceneGraph, SceneNode, NodeType } from './'
+import type { SkiaRenderer } from '@open-pencil/core'
```
dannote: Hit testing now depends on `SkiaRenderer` (and therefore CanvasKit WASM). This breaks the architecture — `scene-graph` is a pure data layer that runs headless in Bun/Node for CLI and tests. Threading a renderer through every hit-test call couples the scene graph to the rendering backend.

The world matrix computation should use pure math (manual 3×3 matrix multiply — it's ~10 lines), not CanvasKit's `Matrix.multiply`. That keeps hit testing independent of WASM.

packages/core/src/scene-graph/index.ts
```diff
@@ -952,12 +965,12 @@ export class SceneGraph {
     this.emitter.emit('node:deleted', id)
   }
 
-  hitTest(px: number, py: number, scopeId?: string): SceneNode | null {
-    return hitTestFn(this, px, py, scopeId)
+  hitTest(px: number, py: number, renderer: SkiaRenderer, scopeId?: string): SceneNode | null {
+    return hitTestFn(this, px, py, renderer, scopeId)
   }
```
dannote, line 953: `hitTest` and `hitTestDeep` signatures now require `SkiaRenderer` — this is a breaking change to the scene graph API. Every consumer (editor, CLI, tests) must now have a renderer available just to do a hit test. See comment on hit-test.ts.

packages/core/src/scene-graph/index.ts
```diff
@@ -1031,4 +1044,73 @@ export class SceneGraph {
     }
     return result
   }
+
```
dannote: `fromJSON` / `getJSON` — unrelated to rotation hit testing, should be a separate PR.

src/views/EditorView.vue
```diff
@@ -37,7 +37,34 @@ const { isMobile } = useViewportKind()
 if (route.meta.demo && !('test' in params)) {
   createDemoShapes(firstTab.store)
 }
+if (route.meta.demo && 'test' in params && process.env.NODE_ENV === 'development') {
```
dannote: `DemoTestStoreData` localStorage persistence and the `Shift+Cmd+Enter` save hotkey are dev-only testing scaffolding — shouldn't be in the PR.

packages/vue/src/shared/input/geometry.ts
```diff
@@ -103,6 +112,116 @@ export function hitTestHandle(
   return null
 }
 
+type HandlePosition = 'nw' | 'ne' | 'se' | 'sw' | 'n' | 'e' | 's' | 'w'
+
+function getCursorAngleFromHandle(handle: HandlePosition, rotation: number): number {
+  const map: Record<HandlePosition, [number, number]> = {
+    nw: [1, 1],
+    ne: [-1, 1],
```
dannote, line 78: This re-declares `HandlePosition` locally — it already exists in `./types`.

packages/vue/src/shared/input/select.ts
```diff
@@ -485,6 +470,8 @@ export function updateHoverCursor(
   editor.setHoveredNode(
     hit && !editor.state.selectedIds.has(hit.id) && hit.id !== editNodeId ? hit.id : null
   )
```
dannote, line 473: This disables hover cursor for multi-selection — intentional behavior change? Needs a comment if so, or removal if not.

packages/vue/src/shared/input/select.ts
```diff
@@ -368,14 +373,17 @@ export function handleSelectDown(
 
   if (editor.state.editingTextId) editor.commitTextEdit()
 
-  const resizeDrag = tryStartResize(sx, sy, cx, cy, editor)
+  if (tryStartRotation(cx, cy)) return
+
+  const resizeDrag = tryStartResize(cx, cy, editor)
```
dannote, line 378: This silently disables resize for multi-selection. Same question — intentional or accidental regression?

dannote (changes requested):
The core idea is valid — `getAbsolutePosition()` sums x/y without accounting for parent rotations, so hit testing and selection overlays break for nested rotated frames. The matrix-based approach is the right fix.

However, this PR needs significant cleanup before merging. See inline comments.

dannote:
Thanks for working on this! Left some comments.

packages/vue/src/shared/input/select.ts
yoohoo360: The multi-select feature currently also has issues on the `master` branch, so I’ll remove the `return` first.

yoohoo360:
All review comments have been addressed and fixed.

dannote:
Merged! Fixed a few remaining non-null assertions (909dc115). Thanks for the contribution 💛

PR #197, opened 2026-04-13
fix(tauri): use opener plugin for external links

sld0Ant:
Fixes #193

---

Clicking external links ("Get an API key →", "Get free Pexels API key →", "Get free Unsplash access key →") in the AI panel triggers `Command plugin:shell|open not allowed by ACL` on Tauri desktop.

**Root cause:** Tauri v2 intercepts `<a target="_blank">` clicks via the shell plugin internally, but only the opener plugin had ACL permissions (`"opener:default"` in capabilities). The `"shell:allow-open"` permission was never granted.

**Fix:** Replace all `<a target="_blank">` with programmatic `openUrl()` from `@tauri-apps/plugin-opener`, which is already whitelisted. A shared `openExternalLink()` composable handles the Tauri/browser switch. Falls back to `window.open()` in the browser.

### What changed
- New `src/composables/use-external-link.ts` — shared helper
- `ProviderSetup.vue` / `ProviderSettings.vue` — `<a>` → `<button @click>`
- E2E test verifying `window.open` is called in browser context

**Checklist:**
- [x] `bun run check` passes (lint + typecheck)
- [x] Tests added or updated
- [x] CHANGELOG.md updated (if user-facing)

PR #198, opened 2026-04-13
fix: ATS exception + logging for WebSocket bridge on macOS

### LINKED ISSUE #190: bug: Tauri release build does not connect WebSocket bridge to MCP server
### Description

  The brew-installed Tauri desktop app (v0.11.4, macOS) does not establish the WebSocket bridge to the MCP server on :7601. Every MCP tool call returns {"error":"OpenPencil app is not
  connected"} because browserRegistered is never set to true on the server side.

  The automation code exists in the source (src/automation/server.ts connects to ws://127.0.0.1:7601 and sends {type:'register', token}), and it works correctly in dev mode (bun run dev). But
  the release binary installed via brew install open-pencil/tap/open-pencil never initiates the WebSocket connection.

  ### Evidence
  - lsof shows zero TCP connections from the OpenPencil process or its WebKit WebContent subprocess to :7601
  - The MCP server's /health endpoint permanently returns {"status":"no_app"} while the Tauri app is running
  - The WebSocket server itself works fine — a Node.js probe can connect, register with the correct auth token, and stay connected
  - Running bun run dev from the repo and opening http://localhost:1420 immediately connects and returns {"status":"ok"} — MCP tools work end-to-end

  Expected: The Tauri app should connect its WebSocket client to the MCP server (either one it spawns itself or an externally-running one), enabling MCP tool calls from Claude Code / other
  agents.

### Proposed fix

  The likely root cause is that macOS WKWebView silently blocks ws:// and http:// connections to localhost from the tauri://localhost origin. In dev mode the app runs in a regular browser tab,
  which doesn't have this restriction.

  The installed app's Info.plist has no NSAppTransportSecurity entry. WKWebView enforces App Transport Security by default, rejecting insecure connections — including the
  fetch('http://127.0.0.1:7600/health') in readHealth() and new WebSocket('ws://127.0.0.1:7601') in connectAutomation().

  Fix 1 (simplest): Add ATS exception to the macOS build

  Create desktop/Info.plist (or add to tauri.macos.conf.json if Tauri v2 supports plist injection) with:
```
  <key>NSAppTransportSecurity</key>
  <dict>
      <key>NSAllowsLocalNetworking</key>
      <true/>
  </dict>
```
  This allows HTTP/WS to 127.0.0.1 and localhost only — no reduction in security for external connections.

  Fix 2 (more robust): Move the bridge to Tauri IPC

  Instead of using browser WebSocket from the webview, create a Tauri command that establishes the WebSocket connection from the Rust process. Tauri plugins bypass WKWebView's network
  restrictions entirely. The health check (readHealth()) could similarly use tauri-plugin-http instead of fetch(). This would make the bridge immune to any webview security policy.

  Fix 3 (diagnostic): Add logging to the connection flow

  Regardless of which fix is chosen, adding console.error logging to the catch blocks in spawnMCPIfNeeded() and the onerror/onclose handlers in connectAutomation() would make this class of
  failure visible instead of silent.

### Steps to reproduce

  1. Install via brew: brew install open-pencil/tap/open-pencil
  2. Start the MCP server: openpencil-mcp (listens on :7600 HTTP, :7601 WS)
  3. Open /Applications/OpenPencil.app
  4. Check health: curl http://127.0.0.1:7600/health — returns {"status":"no_app"}
  5. Any MCP tool call returns OpenPencil app is not connected

  Contrast with dev mode:
  1. bun run dev (spawns MCP server via Vite plugin)
  2. Open http://localhost:1420 in browser
  3. curl http://127.0.0.1:7600/health returns {"status":"ok"}
  4. MCP tools work

### Platform

  macOS (Tauri)

### Browser (web only)

_No response_

### Screenshots

_No response_

sld0Ant:
Fixes #190

## Problem

macOS WKWebView blocks `ws://` and `http://` connections to `127.0.0.1` due to App Transport Security. The release build never connects the WebSocket bridge, so every MCP tool call returns `OpenPencil app is not connected`.

## Changes

- **desktop/tauri.conf.json**: Add `NSAppTransportSecurity.NSAllowsLocalNetworking: true` to macOS bundle — allows insecure localhost connections without affecting external traffic
- **src/automation/spawn-mcp.ts**: Log health check failures in `readHealth()` catch block
- **src/automation/server.ts**: Log WebSocket connect/open/close/error events with `[Automation]` prefix
- **src/components/chat/ProviderSelect.vue**: Log health check retry failures with attempt number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Fix color picker dragging flooding the undo stack — fill/stroke/effect color and opacity drags now collapse into a single undo entry per interaction via debounced batching in `PropertyListRoot`
 - Fix .fig import crash on alias variables without a GUID
+- Fix external links in AI panel blocked by Tauri ACL — use opener plugin instead of shell
 
 ## 0.11.6 — 2026-04-08
 
--- a/desktop/tauri.conf.json
+++ b/desktop/tauri.conf.json
@@ -32,7 +32,12 @@
       "icons/icon.ico"
     ],
     "macOS": {
-      "signingIdentity": "Developer ID Application: Danila Poyarkov (N7D7M3Q3CD)"
+      "signingIdentity": "Developer ID Application: Danila Poyarkov (N7D7M3Q3CD)",
+      "info": {
+        "NSAppTransportSecurity": {
+          "NSAllowsLocalNetworking": true
+        }
+      }
     }
   }
 }
--- a/src/automation/server.ts
+++ b/src/automation/server.ts
@@ -147,12 +147,14 @@ export function connectAutomation(getStore: () => EditorStore, authToken: string
   function connect() {
     try {
       ws = new WebSocket(`ws://127.0.0.1:${AUTOMATION_WS_PORT}`)
-    } catch {
+    } catch (e) {
+      console.error('[Automation] WebSocket constructor failed:', e instanceof Error ? e.message : e)
       scheduleReconnect()
       return
     }
 
     ws.onopen = () => {
+      console.log('[Automation] WebSocket connected to MCP server')
       ws?.send(JSON.stringify({ type: 'register', token }))
     }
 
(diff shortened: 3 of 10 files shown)


dannote:
Cherry-picked the unique changes (ATS fix + logging) into master at 837c09bd — the external-link parts were already merged via #197. Thanks for the contribution 💛

PR #199, opened 2026-04-13
fix: provider settings popover not opening on Windows

### LINKED ISSUE #171: Changing AI doesnt work
### Description

When I want to change AI targeting, I guess with a click on a cog icon, it does nothing. I should be able to change/remove AI model. Now I can not. I am on Windows 11.

### Steps to reproduce

1. Open App
2. Create something
3. Add AI Model
4. Try to remove/change it - Doesn't work

### File

_No response_

### Platform

Windows (Tauri)

### Browser (web only)

_No response_

### Screenshots

Image

sld0Ant:
Fixes #171

## Problem

Clicking the gear icon (⚙️) in the AI chat panel does not open the provider settings popover on Windows 11 (Tauri / WebView2). The popover state changes to `data-state="open"` in the DOM, but it's positioned off-screen (`translate(0px, -200%)`).

## Root Cause

The `<PopoverTrigger>` is wrapped in a `<Tip>` component, which renders a reka-ui `TooltipRoot > TooltipTrigger as-child` around it. Two interactive reka-ui components (Tooltip + Popover) compete for the same DOM element — the tooltip intercepts pointer events before the popover can properly open, and the popper fails to calculate the trigger's position.

This works on macOS WebKit by coincidence due to different event propagation, but fails on Windows WebView2 (Edge-based).

## Fix

Remove the `<Tip>` wrapper from `ProviderSettings.vue` and use a native `title` attribute instead. This eliminates the component conflict — only the Popover handles the trigger element.

### DIFF
--- a/src/components/chat/ProviderSettings.vue
+++ b/src/components/chat/ProviderSettings.vue
@@ -95,14 +95,13 @@ function clearUnsplashKey() {
 
 <template>
   <PopoverRoot>
-    <Tip label="Provider settings">
       <PopoverTrigger
         data-test-id="provider-settings-trigger"
         class="rounded p-0.5 text-muted hover:bg-hover hover:text-surface"
+        title="Provider settings"
       >
         <icon-lucide-settings class="size-3" />
       </PopoverTrigger>
-    </Tip>
 
     <PopoverPortal>
       <PopoverContent


dannote:
Fixed differently in 728274f4 — added a `disabled` prop to `<Tip>` so it force-closes the tooltip when the parent popover opens. This keeps the styled tooltip instead of falling back to a native `title` attribute, and also fixes the same issue in `ColorStyleRow.vue`. Thanks for the contribution 💛

PR #204, opened 2026-04-15
feat: enable headless MCP automation (CLI auto-open + WebKit2GTK race fix + utf-8 stdio)

chrisgliddon:
## User Story

> As an MCP tool author building agent integrations on top of OpenPencil, I want to launch the desktop with a `.fig` path on the command line and immediately drive it via the MCP server, so my agent can run end-to-end tests and design workflows without a human clicking through `File → Open` or restarting the app to recover from a silent injection race.

## Acceptance Criteria

- [ ] Given a fresh shell, when I run `OpenPencil ~/path/to/file.fig`, then the app opens with that file already loaded — no `File → Open` required.
- [ ] Given the app is running on Linux + WebKit2GTK 2.52, when the editor view mounts, then the MCP automation WebSocket connects within 2s on every launch (no "lost the race" failure mode).
- [ ] Given a clean clone of the repo, when I run `bun run build`, then it succeeds without an `oxlint` error blocking the build.
- [ ] Given an MCP client connects to the running app via stdio, when the WebSocket payload arrives as a `Buffer`, then JSON parsing recovers the actual UTF-8 message rather than `"[object Object]"`.

## Problem (today)

Three separate issues block headless agent integrations on Linux:

1. **The MCP automation channel silently fails to connect on WebKit2GTK 2.52.** `IS_TAURI` is computed at module load via `'__TAURI_INTERNALS__' in window`. On WebKit2GTK 2.52 (Linux), Tauri's preload sometimes loses the race with module evaluation, leaving `IS_TAURI = false` for the rest of the session. `spawnMCPIfNeeded` returns `null`, `connectAutomation` is never called, the WS bridge never opens. From the outside the user sees `{"status":"no_app"}` from `/health` and zero diagnostic in the console. Workaround today: close the file and reopen — sometimes works, sometimes doesn't.

2. **Agents can't tell the app what to open.** The editor only learns about a document via `File → Open` (or drop, or share URL). A headless agent has no verb for "boot the app already editing this file" — every automation run starts with a manual click.

3. **`bun run build` fails on a fresh clone.** `packages/mcp/src/stdio.ts:34` does `JSON.parse(String(raw))` where `raw` is a `ws` library payload (`Buffer | ArrayBuffer | Buffer[]`). For some shapes `String(raw)` yields `"[object Object]"` and the JSON parse throws. The same line trips an `oxlint` rule (`typescript-eslint/no-base-to-string`), so `bun run build` exits 1 — which surfaces because `tauri build` invokes `bun run build` as `beforeBuildCommand`. New contributors hit this immediately.

## Solution

Three small, independent commits:

- **`fix(mcp)`** — decode WebSocket payloads explicitly via `Buffer.isBuffer` / `Buffer.concat` / `Buffer.from`. Unblocks `bun run build` and fixes the actual decode bug.
- **`feat(automation)`** — add an async `tauriReady()` helper that returns immediately when `__TAURI_INTERNALS__` is already present, otherwise polls every 50ms for up to 1.5s. `spawnMCPIfNeeded` awaits it instead of reading the module-level `IS_TAURI`. Module-level `IS_TAURI` stays untouched (other call sites that read it synchronously are unaffected).
- **`feat(cli)`** — Rust reads argv on startup; if a `.fig`/`.pen` path is present and exists, it's stashed in a `PendingOpen` Tauri state. New `take_pending_open` command returns + clears the slot exactly once. `EditorView.onMounted` calls it and routes through a new `openFileFromPath` helper that wraps the existing `openFileInNewTab` flow. `capabilities/default.json` `fs:allow-read-file` / `fs:allow-write-file` extended with `$HOME/**` and `/**` so absolute argv paths aren't rejected with "forbidden path" (the existing `File → Open` flow gets an automatic per-session scope grant from `plugin-dialog` that the CLI path doesn't).

## Changes

| File | Change |
|---|---|
| `packages/mcp/src/stdio.ts` | Replace `String(raw)` with `rawToUtf8()` helper that handles `Buffer` / `Buffer[]` / `ArrayBuffer` correctly. |
| `src/utils/tauri-ready.ts` *(new)* | Async helper polling for `__TAURI_INTERNALS__`. |
| `src/automation/spawn-mcp.ts` | `await tauriReady()` instead of reading `IS_TAURI` synchronously. |
| `src/views/EditorView.vue` | Same; also call `take_pending_open` after Tauri is ready and open returned path. |
| `src/composables/use-menu.ts` | Extract `openFileFromPath()` from `openFileDialog()` body; export for reuse. |
| `desktop/src/lib.rs` | `PendingOpen(Mutex<Option<String>>)` state + `take_pending_open` Tauri command + argv parsing. |
| `desktop/capabilities/default.json` | Extend fs scopes with `$HOME/**` and `/**` for absolute argv paths. |

## Why this approach

Considered alternatives:

- **Make `IS_TAURI` itself async/getter.** Would cascade through every existing call site (synchronous reads in `use-menu.ts`, `engine/fonts.ts`, etc.) and break public API. Surgical `tauriReady()` only where it matters keeps the change small and backward-compatible.
- **Open file via Tauri event from Rust → frontend listener.** Tauri events emitted before the listener registers are lost. Polling a state via a command is race-free and consumes the slot exactly once.
- **Read argv as a sidecar CLI plugin.** Tauri's `plugin-cli` exists but is heavyweight for one optional positional argument. A direct `std::env::args()` read in `setup` is 4 lines.

## How to verify

1. `bun install && bun run tauri build --no-bundle` — succeeds without lint error.
2. Launch a standalone `openpencil-mcp` (so an MCP server is reachable on `127.0.0.1:7600`).
3. `OpenPencil ~/some/file.fig` — within ~5s, `curl http://127.0.0.1:7600/health` returns `{"status":"ok", "token":"..."}` and the file is loaded.
4. Connect any MCP client; `tools/list` returns the full 90+ tool surface and calls succeed against the loaded document — no `File → Open` clicked.
5. Repeat the launch 5 times in a row on Linux + WebKit2GTK 2.52 — every launch connects (no race-induced silent failures).

Tested on Arch Linux + KDE Plasma + WebKit2GTK 2.52.1 + Mesa 26 + AMD Radeon Vega.

## Out of scope

- A user-facing settings panel for AI providers / API keys (separate user story; we hit it but didn't try to fix it here).
- Single-instance handoff (`OpenPencil --open-additional foo.fig` against a running instance). Different feature, IPC layer needed.
- Replacing `IS_TAURI` everywhere with the async helper. Most existing call sites are demonstrably fine with the synchronous read; targeted change minimizes blast radius.


dannote (commented):
Thanks for tackling this! The stdio fix is solid. A couple things on the other two:

**`tauriReady()` polling**

Polling `window` every 50ms for 1.5s is a workaround for a problem we should fix at the source. By the time `onMounted` fires, Tauri's preload has always injected `__TAURI_INTERNALS__` — the race is only with module-level evaluation of `IS_TAURI` in `constants.ts`.

The proper fix is to make `IS_TAURI` lazy — either a getter or just check `"__TAURI_INTERNALS__" in window` directly at the two async callsites (`spawnMCPIfNeeded`, `EditorView.onMounted`). No polling needed.

**FS scope in capabilities**

`{ "path": "/**" }` grants read/write to the entire filesystem. That's a significant security regression — the current `**` scope is relative and gets narrowed by `plugin-dialog` per-session grants. For the CLI open path, a better approach would be to use Tauri's `fs:allow-read-file` with a runtime scope grant for the specific argv path, rather than opening up everything.

dannote:
Both review points from the previous round are fully addressed — thanks for the quick turnaround.

**Resolved:**
- `tauriReady()` polling removed, inline `'__TAURI_INTERNALS__' in window` checks are clean.
- Broad FS scope reverted; Rust eagerly reads argv bytes and passes them through `take_pending_open` — no frontend filesystem access needed.

**New issue — `src/engine/google-fonts.ts`:**

This file is 189 lines of dead code (never imported anywhere in the PR) that duplicates existing functionality in `packages/core/src/text/fonts.ts`:

- `styleToVariant()` — identical copy
- `styleToWeight()` — identical copy
- `FontFace` registration — `registerFontInBrowser()` already does this
- Google Fonts fetching — `fetchGoogleFont()` already does this

The PR description doesn't mention fonts. Please remove this file from the PR. If IndexedDB font caching is a feature you want to add, open a separate PR against `packages/core/src/text/fonts.ts` instead of introducing a parallel implementation.

Everything else looks good to merge once that's removed.

dannote:
Thanks again for the original investigation and implementation. I extracted the remaining useful part and committed it directly to master in ddf415ec: desktop file associations / Open With support for .fig and .pen, with single-instance handoff for already-running apps.

The MCP UTF-8 decoding and Tauri detection pieces were already fixed on master, and the unrelated font changes were left out, so this PR is now obsolete.

dannote:
Closing as obsolete. The remaining useful functionality has been landed on master in ddf415ec; the other fixes are already present.

Issue #208, opened 2026-04-15
DX improvements from MCP automation stress test (5 findings)


chrisgliddon:
## User Story

> As a tool author building a design-system pipeline on top of OpenPencil's MCP, I want the editor + MCP surface to be predictable and self-describing enough that my agent can author production-quality components without surprise gaps, so I can ship a real workflow without having to fork the editor first.

## Acceptance Criteria (one per finding)

- [ ] Bundled font set covers at least the common categories a design tool needs: a default sans, a default mono, a default display/pixel face (or a clear path to import system fonts).
- [ ] The JSX renderer supports a notion of overlay / absolute positioning inside an auto-layout container, so designers can put a badge in the corner of a card without breaking the parent's layout flow.
- [ ] An icon set beyond Iconify Lucide is available, OR a clear way to import custom SVG icons that survive `save_file` (related: #vector-tangents issue).
- [ ] The in-app AI panel ("Claude Code", "Codex", "Gemini CLI") has a Settings UI for selecting / configuring the provider — not just a dropdown that silently fails on send.
- [ ] When the desktop app isn't connected to the MCP server, `tools/list` either gates the surface or annotates it so clients can show "0 tools, app not connected" upfront instead of failing per-call.

## Findings (severity-ordered)

These came out of a single ~1h stress-test session driving the MCP surface to build a real component (a card with image area, gradient grade bar, mana bar, wear meter, keyword chips, drop shadow, stroke, variable bindings) plus a 3×5 combat-HUD layout, a custom vector path, mixed-color text via `set_font_range`, and a component → instance flow.

The 2 P0/P1 bugs from the same session are filed separately:

- save_file vector tangent crash — see #206 / #207
- node_move string-coercion — see #206 / #207

This issue bundles the 5 lower-severity DX gaps so the maintainer has one place to triage.

---

### 1. Only Inter is bundled as a font

**Symptom:** `list_fonts` returns `{count: 1, fonts: [{family: "Inter", weights: [400, 700]}]}`. Setting any other family on a text node fails or falls back silently.

**Why it matters:** Every real design language needs at least a sans + mono + a display face. Inter alone forces every project to either ship custom font assets or live with one face throughout. Particularly painful for game UI, retro/pixel design, code-adjacent tooling.

[...]

PR #210, opened 2026-04-15
feat(mcp): add list_available_fonts tool for font discovery

chrisgliddon:
## User Story

> As an MCP tool author writing an agent that authors new text content in OpenPencil, I want to discover which font families the host can actually render — not just the families already in use — so I can pick a real font with confidence instead of trial-and-erroring against \`set_font\`.

## Acceptance Criteria

- [ ] Calling \`list_available_fonts\` from any MCP client returns the same family list a user would see in the Properties panel font picker.
- [ ] The optional \`family\` filter does a case-insensitive substring match (matching \`list_fonts\` UX).
- [ ] In non-desktop / test runtimes that don't expose enumeration, the tool returns \`{count: 0, fonts: [], note: ...}\` rather than throwing.
- [ ] No regression to the existing \`list_fonts\` tool (which intentionally remains scoped to *currently used* fonts on the page).

## Problem (today)

\`list_fonts\` (\`packages/core/src/tools/read.ts:198\`) walks every \`TEXT\` node on the current page and groups by \`fontFamily\`:

> *List fonts used in the current page.*

That's exactly the right contract for a "what's in this design" inspection — but it gives agents zero way to *discover* what they could set. The desktop already enumerates system fonts via the \`list_system_fonts\` Tauri sidecar (consumed by \`engine/fonts.ts#listFamilies\` and rendered in the Properties panel font picker), but no MCP tool exposes that list. So an agent's only option is to guess a family name and hope \`set_font\` succeeds.

Concrete failure case: a fresh document containing only one text node with Inter. \`list_fonts\` returns \`Inter\`. A coding agent reasoning "what fonts can I use?" sees nothing else and either (a) sticks with Inter forever, or (b) guesses families that may or may not exist on the host.

## Solution

Add a sibling \`list_available_fonts\` tool that mirrors what the Properties-panel font picker shows. The two tools coexist — same shape, different scope:

| Tool | Returns | Use case |
|---|---|---|
| \`list_fonts\` | Families currently used on the page | Inspecting an existing design |
| \`list_available_fonts\` | Families the host can render | Picking a font for new text |

## Changes

| File | Change |
|---|---|
| \`packages/core/src/figma-api/index.ts\` | Add optional \`listAvailableFontFamilies?: () => Promise<string[]>\` to \`FigmaAPI\`. Same opt-in pattern as the existing \`exportImage?\` field. |
| \`packages/core/src/tools/read.ts\` | Define the new tool. Returns \`{count, fonts, note?}\`. Falls back gracefully when the host doesn't implement the hook. |
| \`packages/core/src/tools/registry.ts\` | Add to \`CORE_TOOLS\` so the MCP server advertises it. |
| \`src/automation/figma-factory.ts\` | Wire the desktop implementation by delegating to \`engine/fonts.ts#listFamilies\`, which already exists and already calls the \`list_system_fonts\` Tauri sidecar. |

## Why this approach

Considered alternatives:

- **Make \`list_fonts\` itself toggleable via a \`scope: 'page' | 'available'\` parameter.** Would conflate two distinct user intents into one tool's response shape (the existing tool returns weights per family; an availability listing is family-only). Two tools with one job each is clearer for both human readers and tool-using agents than one tool with a mode flag.
- **Expose via a custom command handler in \`src/automation/server.ts\` (alongside \`save_file\`, \`export\`, etc.).** Would skip touching core, but the MCP server then can't advertise it via \`tools/list\` because the registry comes from \`CORE_TOOLS\`. Following the existing tool-definition pattern keeps everything discoverable and consistent.
- **Bundle a default font set so \`list_fonts\` is meaningful even on a fresh doc.** Worth doing separately (and tracked in #208 finding 1) but is a much larger change. This PR is the minimal discovery gap fix.

The \`exportImage?\` precedent (added in an earlier PR for the same reason — desktop-only capability surfaced through the FigmaAPI proxy with graceful fallback in non-desktop hosts) is the model this follows.

## How to verify

1. Build + launch the desktop on any platform with system fonts installed.
2. Connect any MCP client. Confirm \`tools/list\` includes \`list_available_fonts\`.
3. Call \`list_available_fonts({})\` — should return a multi-dozen-family list matching what's in the Properties panel font picker.
4. Call \`list_available_fonts({family: "mono"})\` — should return a filtered list of mono-named families.
5. Call \`set_font({id: <text node>, family: <a family from the response>, style: "Regular", size: 16})\` — should succeed and render in the chosen face.
6. Call \`list_fonts({})\` — confirm unchanged (only families used on the page).

Built and lint-clean on Arch Linux + WebKit2GTK 2.52 + Mesa 26 + AMD Radeon Vega.

## Out of scope

- **Bundling default fonts.** Discussed in #208 finding 1; warrants its own conversation about binary-size impact + license curation. This PR fills the discovery gap regardless of whether the host bundles fonts or not.
- **Per-family weight/style listing.** \`listFamilies()\` returns family names only today. Returning \`{family, weights, styles}\` per entry would be valuable but requires plumbing through the \`list_system_fonts\` sidecar's existing weight enumeration; happy to do as a follow-on if maintainers want it in this PR or separately.
- **Caching.** \`engine/fonts.ts\` already memoizes the Tauri call; the tool inherits that without additional code.

### DIFF
--- a/oxlint.json
+++ b/oxlint.json
@@ -38,6 +38,7 @@
             "allowImportNames": [
               "loadFont",
               "listFamilies",
+              "listFonts",
               "preloadFonts"
             ]
           },
--- a/packages/core/src/figma-api/index.ts
+++ b/packages/core/src/figma-api/index.ts
@@ -1,7 +1,14 @@
 import { IS_BROWSER } from '../constants'
 import { computeBounds } from '../geometry'
 import { copyFills, copyStrokes, copyEffects } from '../scene-graph/copy'
-import { FigmaNodeProxy, INTERNAL_ID, MIXED, type FigmaFontName, type NodeProxyHost } from './proxy'
+import {
+  FigmaNodeProxy,
+  INTERNAL_ID,
+  MIXED,
+  type FigmaFont,
+  type FigmaFontName,
+  type NodeProxyHost
+} from './proxy'
 
 import type { RasterExportFormat } from '../io/formats/raster'
 import type {
@@ -15,7 +22,7 @@ import type {
 import type { Rect, Vector } from '../types'
 
 export { FigmaNodeProxy } from './proxy'
-export type { FigmaFontName } from './proxy'
+export type { FigmaFont, FigmaFontName } from './proxy'
 
 export function computeImageHash(data: Uint8Array): string {
   let h1 = 0x811c9dc5 >>> 0
@@ -370,6 +377,12 @@ export class FigmaAPI implements NodeProxyHost {
     // No-op: we don't gate text editing on font loading
   }
 
+  async listAvailableFontsAsync(): Promise<FigmaFont[]> {
+    // Default: pure browser / test contexts have no enumeration surface.
+    // Desktop hosts override this to return system + bundled fonts.
+    return []
+  }
+
   notify(message: string): { cancel: () => void } {
     if (typeof console !== 'undefined') console.log(`[figma.notify] ${message}`)
     // eslint-disable-next-line no-empty-function
@@ -385,12 +398,4 @@ export class FigmaAPI implements NodeProxyHost {
     nodeIds: string[],
     options: { scale?: number; format?: RasterExportFormat; quality?: number }
   ) => Promise<Uint8Array | null>
-
-  /**
-   * Returns every font family the host can render — system fonts on
-   * desktop (via font-kit) plus any bundled fonts the host registers.
-   * Optional because pure browser/test contexts have no enumeration
-   * surface; tools that depend on this should fall back gracefully.
-   */
-  listAvailableFontFamilies?: () => Promise<string[]>
 }
--- a/packages/core/src/figma-api/index.ts
+++ b/packages/core/src/figma-api/index.ts
@@ -385,4 +385,12 @@ export class FigmaAPI implements NodeProxyHost {
     nodeIds: string[],
     options: { scale?: number; format?: RasterExportFormat; quality?: number }
   ) => Promise<Uint8Array | null>
+
+  /**
+   * Returns every font family the host can render — system fonts on
+   * desktop (via font-kit) plus any bundled fonts the host registers.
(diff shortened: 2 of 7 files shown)


dannote (commented):
Thanks for the contribution!

One thing — `FigmaAPI` mirrors the Figma Plugin API, and Figma already has exactly this:

```ts
figma.listAvailableFontsAsync(): Promise<Font[]>
// Font = { fontName: { family: string, style: string } }
```

So instead of the custom `listAvailableFontFamilies` property, this should be implemented as `listAvailableFontsAsync()` on the `FigmaAPI` class — same name, same shape. The tool can dedupe families from the result. That way `eval` and any future plugin-compat code also get it for free.

We already have `loadFontAsync` on `FigmaAPI` following this pattern.

Could you update to match the Figma API shape?

chrisgliddon:
Thanks for the review — you're right, mirroring the Figma plugin API is the right call. Pushed [`f53e0bc`](https://github.com/open-pencil/open-pencil/pull/210/commits/f53e0bc) reshaping to match:

- `FigmaAPI.listAvailableFontsAsync(): Promise<Font[]>` — real method with a safe `[]` default, same pattern as `loadFontAsync`. Dropped the optional `listAvailableFontFamilies` hook entirely.
- `Font = { fontName: { family, style } }` exported from `figma-api/proxy.ts` alongside `FigmaFontName`.
- Tool dedupes families from the result via a `Set` — public shape of `list_available_fonts` is unchanged (still `{count, fonts: string[]}`).
- Desktop factory flattens `{family, styles[]}[]` → `Font[]`, so `eval` now sees one entry per `(family, style)` pair — e.g. separate `Arial Regular`, `Arial Bold`, `Arial Italic`.

`oxlint --type-aware --type-check` + `vue-tsc --noEmit` both clean.

dannote:
Merged, thanks for the quick turnaround on the review feedback 💛

PR #215, opened 2026-04-20
fix(mcp): surface eval return value and improve stdio reliability

dymo:
## Summary

Two small fixes for pain points surfaced during real-world agent usage:

### 1. eval: surface last expression value (REPL-style)

Previously, code ending with an expression statement (e.g. `JSON.stringify(results)`) returned `undefined` — the IIFE wrapper discarded it. The agent had no way to know the call returned a value and was forced to issue a follow-up query.

`wrapEvalCode()` now detects when the last non-empty line looks like a value expression (not a statement keyword or closing brace) and wraps it in `return (expr)`, matching REPL behaviour. Code that explicitly starts with `return` is used verbatim.

### 2. stdio: clearer error and mcpRoot wiring

- The "app not connected" error now instructs the AI to **stop and
  surface the problem to the user** rather than trying to resolve it
  autonomously (mitigates issue #208 point 5)
- `OPENPENCIL_MCP_ROOT` is now read in `stdio.ts` and passed to
  `registerTools` — it was missing, causing a crash on startup when the
  env var was unset

## Testing

- Add 17 unit tests covering output semantics and wrapping shape (`tests/engine/eval-wrap.test.ts`)

### DIFF
--- a/packages/mcp/src/stdio.ts
+++ b/packages/mcp/src/stdio.ts
@@ -77,7 +77,12 @@ function connect() {
 function sendRpc(body: Record<string, unknown>): Promise<unknown> {
   return new Promise((resolve, reject) => {
     if (!ws || ws.readyState !== WebSocket.OPEN || !registered) {
-      reject(new Error('OpenPencil app is not connected. Start the app and open a document.'))
+      reject(new Error(
+        'OpenPencil app is not connected. ' +
+        'STOP and tell the user: "The OpenPencil desktop app is not running or no document is open. ' +
+        'Please start the app and open a document, then try again." ' +
+        'Do NOT attempt to start the app yourself or retry automatically.'
+      ))
       return
     }
     const id = crypto.randomUUID()
--- a/src/automation/eval-wrap.ts
+++ b/src/automation/eval-wrap.ts
@@ -1,0 +1,32 @@
+const STATEMENT_KEYWORDS =
+  /^(const|let|var|function|class|if|for|while|do|switch|try|throw|return|break|continue|import|export|async\s+function|\/\/|\/\*)/
+
+/**
+ * Wrap eval code so the last bare expression is returned (REPL-style).
+ *
+ * - Already starts with `return` → used verbatim
+ * - Last non-empty line looks like a statement keyword → wrap in IIFE so we
+ *   can still await async side-effects
+ * - Otherwise → split off the last line and promote it to `return (lastLine)`
+ */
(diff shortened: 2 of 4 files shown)


src/automation/eval-wrap.ts
```diff
@@ -0,0 +1,32 @@
+const STATEMENT_KEYWORDS =
```
dannote, line 1: The regex-based heuristic is fragile. `42;` returns undefined but `42` returns 42 — no REPL works that way. Other edge cases: `someObj.class` false-positives on `class`, multi-line expressions, `{a: 1}` ambiguity.

We already have `sucrase` as a dependency — use its parser to find the last expression statement in the AST instead of pattern-matching lines.

src/automation/server.ts
```diff
@@ -233,6 +232,33 @@ export function connectAutomation(getStore: () => EditorStore, authToken: string
...
 }
 
+/**
+ * Wrap eval code so the last expression value is returned, REPL-style.
+ * If the code already starts with `return`, use it verbatim.
+ * If the last non-empty line looks like a value expression (not a statement
+ * keyword or closing brace), promote it to `return (expr)` so the result
+ * is surfaced to the caller instead of being silently discarded.
+ */
+function wrapEvalCode(code: string): string {
```
dannote, line 242: This is a second, different implementation of `wrapEvalCode` — dead code since the import from `eval-wrap.ts` on line 23 shadows it. Remove.

packages/mcp/src/stdio.ts
```diff
@@ -77,7 +77,12 @@ function connect() {
 function sendRpc(body: Record<string, unknown>): Promise<unknown> {
   return new Promise((resolve, reject) => {
     if (!ws || ws.readyState !== WebSocket.OPEN || !registered) {
-      reject(new Error('OpenPencil app is not connected. Start the app and open a document.'))
+      reject(new Error(
+        'OpenPencil app is not connected. ' +
+        'STOP and tell the user: "The OpenPencil desktop app is not running or no document is open. ' +
```
dannote, line 82: This overlaps with #216 which has the same error message change and `mcpRoot` wiring. Coordinate to avoid merge conflicts.

dannote (changes requested):
Thanks for working on this!

dannote:
The REPL-style eval return and stdio improvements landed in 9e013e60. Single wrapEvalCode() implementation in core, used by both the tool and automation bridge. Thanks for the contribution!

PR #217, opened 2026-04-20
feat(mcp): write export output to disk via optional `path` parameter

dymo:
## Summary

`export_image`, `export_svg`, and `get_jsx` each gain an optional `path` parameter. When provided and `OPENPENCIL_MCP_ROOT` is set, the MCP server writes the output directly to disk and returns `{ written, byteLength }` instead of the raw data.

This avoids flooding the AI context window with large base64 blobs or long JSX strings when the agent only needs to know the file was produced.

## Changes

- Add optional `path` to `export_image`, `export_svg`, and `get_jsx` tools
- Server validates the path against `OPENPENCIL_MCP_ROOT` and returns `{ written, byteLength }` on success
- Requires `OPENPENCIL_MCP_ROOT` to be set (paths outside the root are rejected)

## Notes

- Path validation reuses the same `resolvedRoot` + `osSep` check from PR #215 — requires `OPENPENCIL_MCP_ROOT` to be set
- No changes to tool execution in the browser; writing happens entirely in the Node.js MCP server process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,21 @@
 ### Features
 
 - Add stdio transport for MCP server — `openpencil-mcp` now works as a proper stdio MCP server for Claude Code, Cursor, etc. HTTP server available as `openpencil-mcp-http`.
+- Add `open_file` MCP tool — opens a .fig or .pen file from disk into a new tab without user interaction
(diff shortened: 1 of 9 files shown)


dannote:
Merged into master with cleanup — deduplicated path validation into `resolveSafePath`, reused `openFileFromPath` instead of reimplementing, kept the stdio relay simple. Thanks for the great feature! 💛

PR #222, opened 2026-04-21
Dont mind me just testing out your preview workflow...

joeycumines:


joeycumines:
@dannote @sld0Ant Please fix the preview workflow. As-is the build executes with write access to PRs etc, and the preview mechanism uses your prod pages project - not ideal.

N.B. Your cloudflare token is likely still secure.

PR #230, opened 2026-04-23
Fix Windows MCP bridge spawn

LCbeijing:
## Summary
- make desktop automation bridge startup more robust on Windows by launching the npm wrapper via cmd /d /c
- add a small helper so the platform-specific command path is explicit and testable
- allow the Windows command alias in desktop shell capabilities

## Testing
- bun test tests/engine/spawn-mcp.test.ts
- bun test tests/engine/mcp-stdio.test.ts


dannote:
Already fixed in v0.11.8 — we landed the same `cmd /c` approach in `src/app/automation/mcp/spawn.ts`. Thanks for flagging it!

PR #231, opened 2026-04-23
Support MCP root saves on Unicode Windows paths

LCbeijing:
## Summary
- route MCP-root saves through the host filesystem instead of Tauri file writes
- add an automation command that exports the current .fig bytes directly from the live document
- make 
ew_document(path) and save_file(path) work in MCP-root workflows even when Windows desktop file APIs struggle with Unicode paths

## Why
On Windows, the desktop/Tauri file-writing path can fail for MCP automation when the target path contains non-ASCII characters. This patch keeps the design automation flow in the live app, but performs the actual file write in the MCP host process, which already handles Unicode paths correctly.

## Testing
- bun test tests/engine/mcp-server.test.ts

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -47,6 +47,12 @@ export function fail(e: unknown): MCPResult {
 
 export type RpcSender = (body: Record<string, unknown>) => Promise<unknown>
 
+interface RpcEnvelope {
+  ok?: boolean
+  result?: unknown
+  error?: string
+}
+
 export interface RegisterToolsOptions {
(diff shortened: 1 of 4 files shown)


dannote:
Thanks for digging into this and for the clear Windows Unicode repro. We ended up landing a narrower fix on master that keeps save ownership in the app/Tauri document IO path while using scoped MCP paths and native dirname handling for Unicode Windows paths, so the host-side fig export/write path here isn’t needed anymore. I’m going to close this out, but the underlying issue should be covered now.

PR #238, opened 2026-04-30
fix(vector): render dashPattern strokes correctly on closed crescents

greekr4:
## Summary

Vector nodes with `stroke.dashPattern` were rendering as solid lines, and even when the dash pattern was applied through brute outline-stroke conversion, closed crescent shapes (e.g. annular wedges) showed two parallel dashed arcs instead of a single dashed arc along the centerline. After the fix, a dashed `<Vector>` renders the same way as in Figma.

## Repro

```jsx
<Vector
  name="Arc"
  x={28} y={26}
  w={227} h={116}
  stroke="#7A7A7A"
  strokeDash={[6, 6]}
/>
```

## Screenshots

**Figma (source of truth):**

figma-original

**open-pencil before — pasted closed crescent renders as two parallel solid arcs (no dash):**

before

**open-pencil after — single smooth dashed arc along the centerline, matching Figma:**

after

## Root cause

Two independent bugs combined.

1. `drawVectorPathStrokes` in `packages/core/src/canvas/scene.ts` converts the stroke into a fill outline (`vp.copy().stroke(strokeOpts)`) and then draws that outline with `fillPaint`. There is no place left for `PathEffect.MakeDash` to apply, so the dash never appears.
2. A closed crescent (e.g. `annularWedge`) is one closed cycle that visits the outer arc and then the inner arc. Even when the dash effect is applied to the raw path, both arcs receive dashes, producing two parallel dashed arcs.

### 1. `packages/core/src/canvas/scene.ts`

- `drawVectorPathStrokes`: when `stroke.dashPattern` is non-empty, skip the outline conversion and call `canvas.drawPath` with `strokePaint` configured with `PathEffect.MakeDash`. This keeps the original (non-dashed) outline path for normal strokes.
- `renderShapeUncached`: for dashed `VECTOR` nodes that have a `vectorNetwork`, route through the new centerline path instead of the precomputed `strokeGeometry`. The precomputed geometry is already a fill outline and cannot be dashed.

### 2. `packages/core/src/vector/index.ts` — `vectorNetworkToCenterlinePath`

For closed thin crescents:

1. **Detection** (`isClosedThinCrescent`): vertex count `n >= 6 && n % 2 === 0`, every vertex has degree 2, and the cycle has small thickness — `pairedAvg <= alongSum * 0.5` and the paired distances are consistent (`stdDev / pairedAvg <= 0.5`). This rejects rectangles, hexagons, and other regular polygons.
2. **Cap detection by direction change**: along an arc, segment-to-segment turns are smooth (~7.5°); at a cap they are sharp (~90°). A vertex is a "corner" if the angle between its incoming and outgoing segment directions exceeds 60°. A cap segment has both endpoints corners.
3. **Subchain split**: with the two cap indices known, walk the cycle from one cap to the other to get two subchains (outer arc, inner arc). The two chains run in opposite directions around the cycle; reversing one aligns vertices by angle.
4. **Centerline midpoints**: pair `chainA[i]` with `chainB[len-1-i]` and take midpoints. For an `annularWedge(rIn=80, rOut=100)` these land exactly on the midradius=90 circle.
5. **Smooth arc**: `fitCircleArc` runs a 3-point circle construction on first/middle/last midpoints, verifies all midpoints lie on the resulting circle within tolerance, and returns `{cx, cy, r, startAngleDeg, sweepDeg}`. The path is then a single `Path.addArc` call — no polyline kinks. If the midpoints are not co-circular, the function falls back to a midpoint polyline.

For open paths and other shapes the function uses the existing chain walk via `buildChains`, so dashing an open arc continues to work.

### 3. JSX prop wiring (so users can author dashed vectors)

- `packages/core/src/design-jsx/tree.ts`: add `strokeDash?: number[] | boolean` to `StyleProps`.
- `packages/core/src/design-jsx/renderer.ts`: new `applyStrokeOverrides` that delegates to `parseStroke` and then assigns `stroke.dashPattern` from `resolveDashPattern(props.strokeDash, strokeWidth)`. `true` expands to `[w*2, w*2]` (clamped to a min of 4); arrays are filtered to positive finite numbers.
- `packages/core/src/io/formats/jsx/export.ts`: `solidStroke` now also returns `dash`, and `collectAppearanceProps` emits a `strokeDash` prop when present, so JSX export round-trips dashed strokes.

### 4. Editor UI — `src/components/properties/StrokeSection.vue`

Adds a dash toggle button plus dash-length and gap `ScrubInput`s. The toggle stores `[]` when off and `[dash, gap]` when on, matching the `Stroke.dashPattern` model. Helper functions `dashState`/`toggleDash`/`setDash`/`setGap` keep the template thin. No localized strings — icon-only, consistent with existing rows.

## Verification

- `bun run check` passes (oxlint, tsgo, vue-tsc). Two `max-lines` warnings are pre-existing for `scene.ts` (706 lines, was already over 600) and `vector/index.ts` (632 lines after additions).
- `bun test ./tests/engine` shows 3 failing tests on this branch — all 3 also fail on a clean `upstream/master` checkout and are unrelated to this change (`drop shadow has TEXT-specific branch`, `drop shadow uses spread for shape expansion`, `imported nested instance layout recomputes hidden sibling offsets`).
- Visually verified against the Figma original for `<Vector strokeDash={[6,6]}>` rendering of a closed crescent (semicircle annular wedge) — single smooth dashed arc along the centerline, matching Figma.

## Approaches that were tried and rejected

| Approach | Why it failed |
| --- | --- |
| Outline conversion then dash | Outline is a fill — `PathEffect.MakeDash` does not apply. |
| Outline kept, dash on raw path | Closed crescent dashes both outer and inner arc — visible as two parallel dashed arcs. |
| Use the two shortest segments as caps | Fails on thick crescents where `R - r > pi*r/n` and the cap can be the longest segment. |
| Centroid-based clustering | The centroid of an arc is not at the arc center, so cluster boundaries land in the wrong place. |
| Closest non-adjacent vertex pairing | When outer-step is shorter than crescent thickness, the partner becomes an adjacent same-arc vertex. |
| Use the outer subchain itself | Result sits on the outer edge; the user expects the centerline. |
| Midpoint polyline (`lineTo`) | ~7.5° kinks per segment make the arc look like a polygon. |

Direction-change cap detection plus 3-point circle fit is the only approach that holds across thin/thick crescents and synthetic/real shapes.

## Files

- `packages/core/src/canvas/scene.ts`
- `packages/core/src/vector/index.ts`
- `packages/core/src/design-jsx/tree.ts`
- `packages/core/src/design-jsx/renderer.ts`
- `packages/core/src/io/formats/jsx/export.ts`
- `src/components/properties/StrokeSection.vue`

### DIFF
--- a/packages/core/src/canvas/scene.ts
+++ b/packages/core/src/canvas/scene.ts
@@ -1,4 +1,5 @@
 import { DROP_HIGHLIGHT_ALPHA, DROP_HIGHLIGHT_STROKE, SECTION_CORNER_RADIUS } from '../constants'
+import { vectorNetworkToCenterlinePath } from '../vector'
 
 import type { SceneNode, SceneGraph } from '../scene-graph'
 import type { Color } from '../types'
@@ -352,6 +353,19 @@ function drawVectorPathStrokes(
   stroke: SceneNode['strokes'][0],
   sc: Color
 ): void {
+  const dash = stroke.dashPattern
+  if (dash && dash.length > 0) {
+    r.strokePaint.setColor(r.ck.Color4f(sc.r, sc.g, sc.b, sc.a))
+    r.strokePaint.setAlphaf(stroke.opacity)
+    r.strokePaint.setStrokeWidth(stroke.weight)
+    r.strokePaint.setStrokeCap(getCapEntity(r, stroke.cap ?? 'NONE'))
+    r.strokePaint.setStrokeJoin(getJoinEntity(r, stroke.join ?? 'MITER'))
+    r.strokePaint.setShader(null)
+    r.strokePaint.setPathEffect(r.ck.PathEffect.MakeDash(dash, 0))
+    for (const vp of vectorPaths) canvas.drawPath(vp, r.strokePaint)
+    r.strokePaint.setPathEffect(null)
+    return
+  }
   const strokeOpts = {
     width: stroke.weight,
     miter_limit: 4,
@@ -472,6 +486,18 @@ export function renderShapeUncached(
     const stroke = node.strokes[si]
     if (!stroke.visible) continue
     const sc = r.resolveStrokeColor(stroke, si, node, graph)
+
+    if (
+      stroke.dashPattern &&
+      stroke.dashPattern.length > 0 &&
+      node.type === 'VECTOR' &&
+      node.vectorNetwork
+    ) {
+      const centerline = vectorNetworkToCenterlinePath(r.ck, node.vectorNetwork)
+      drawVectorPathStrokes(r, canvas, [centerline], stroke, sc)
+      centerline.delete()
+      continue
+    }
 
     drawNodeStroke(r, canvas, node, rect, hasRadius, stroke, sc, sg, vectorPaths, vectorStroke)
   }
--- a/packages/core/src/design-jsx/renderer.ts
+++ b/packages/core/src/design-jsx/renderer.ts
@@ -79,6 +79,27 @@ function parseStroke(value: string, width: number): Stroke {
   }
 }
 
+function resolveDashPattern(value: unknown, strokeWidth: number): number[] | null {
+  if (Array.isArray(value)) {
+    const out = value.map((n) => Number(n)).filter((n) => Number.isFinite(n) && n > 0)
+    return out.length > 0 ? out : null
+  }
+  if (value === true) {
+    const d = Math.max(Math.round(strokeWidth * 2), 4)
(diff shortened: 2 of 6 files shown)


dannote:
Really nice contribution — you clearly went deep on this one and it shows. I rebased on latest master and pushed a small follow-up (centerline extraction to its own module, PathEffect leak fix, aria-label, unit tests). Merging! 💛

PR #245, opened 2026-05-04
Fixed the issue of Chinese fonts not displaying on UI drafts.

Tem-man:
packages/core/src/text/fonts.ts

The `findLocalFont` function adds the optional parameter `allowVariable`. Using `allowVariable: true` on the system CJK candidate path in `ensureCJKFallback` allows registration of native variable CJK fonts, preventing them from being skipped incorrectly.

Regardless of local success, `CJK_GOOGLE_FONTS` (Noto Sans SC/JP/KR) will be tried again in `Promise.allSettled`, merged with the local list to remove duplicates, and complete the glyphs and fallback chain.

packages/core/src/constants.ts

Microsoft YaHei UI and SimSun have been added as Windows candidates to improve the probability of native matching.

### DIFF
--- a/packages/core/src/canvaskit.ts
+++ b/packages/core/src/canvaskit.ts
@@ -1,3 +1,4 @@
+/// <reference types="vite/client" />
 import CanvasKitInit, { type CanvasKit } from 'canvaskit-wasm'
 
 import { IS_BROWSER } from './constants'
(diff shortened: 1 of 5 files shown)


dannote (approved):
Thanks for the fix — I cleaned up the formatting/package-manager churn and kept the PR focused on the CJK font fallback and CanvasKit asset handling changes.

Tem-man:
Thank you

PR #252, opened 2026-05-08
fix(mcp): fix protocol mismatch between stdio-bridge and server

stiff:
Fixes protocol mismatch between stdio-bridge and server:
- Stdio-bridge was waiting to RECEIVE 'register' from server, 
- while the server expected to RECEIVE 'register' from client
- neither was sending,
- causing stdio-bridge to block forever.

Changes:

- packages/mcp/src/server.ts:
  - Send { type: 'register', token } to clients on WebSocket open
  - Handle 'request' from stdio-bridge, forward to browser automation
  - Route 'response' from automation back to correct client by requestId
  - Track request queue with per-connection handlers for multiple
    clients
  - Clean handler fallback before client registers (returns error)
  
---

**Checklist:**
- [ ] Tests added or updated

### DIFF
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -1,0 +1,16 @@
+# @open-pencil/mcp
+
+MCP (Model Context Protocol) server for OpenPencil browser automation.
+
+## Architecture
+
+```
+┌─────────────┐     ┌──────────────────┐     ┌──────────────────┐
+│ Agent Apps  │     │   Browser RPC    │     │  Browser App     │
+│ (via stdio) │────▶│   Bridge (port   │◀────│  (automation)    │
+└─────────────┘     │   7601)          │     └──────────────────┘
+                    │                  │
+                    │  Multi-client    │
+                    │  connection hub  │
+                    └──────────────────┘
+```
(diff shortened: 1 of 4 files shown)


dannote:
The core fix from this PR (server sending register token on WebSocket connect) landed in 9e013e60. Cleaned up and merged with the eval fix from #215. Thanks for identifying the deadlock!

stiff:
The core fix is omitted, sending only the token doesn't help events reaching automation server. Look at [packages/mcp/README.md](https://github.com/open-pencil/open-pencil/pull/252/changes#diff-ba5dea79fe8446f1a9461322ff9b3158efbf2e11e1bf709973a2c146fba55539R1) in this PR

I attached script to make few bridge calls as agents do, but without the need for external agents.
[stdio-demo.ts](https://github.com/user-attachments/files/27595629/stdio-demo.ts)

Probably it would make sense to convert it into proper test?

dannote:
Thanks again for the PR and for pushing on the repro. I ended up landing the same routing fix in v0.12.2, with the WebSocket routing centralized in `browser-rpc.ts` and tests for stdio requests, late desktop registration, reconnects, and disconnected responses. Really appreciate the debugging here — it identified the actual mismatch.

stiff:
Now works as expected and cleaner than what I did (got confused with many RPCs)

PR #253, opened 2026-05-10
update id and name of deepseek as deepseek-v4-flash and deepseek-v4-pro

Tem-man:
image

Only the file was modified  -> packages/core/src/constants.ts
The specific modifications are as follows:
	models: [
			{ id: "deepseek-v4-flash", name: "DeepSeek-v4-Flash", tag: "Fast" },
			{ id: "deepseek-v4-pro", name: "DeepSeek-v4-Pro", tag: "Reasoning" },
		],

suggestion to update @ai-sdk/deepseek to 2.0.34


dannote (approved):
Thanks! I fixed the default model and SDK update in a follow-up commit.

PR #256, opened 2026-05-13
fix: image fills lost after save . hex hash corrupted in Kiwi byte[] encoding

guimaurer:
## Problem

Image fills disappear after saving a `.fig` file. When reopening or after the file watcher reloads the document, rectangles with image fills render as transparent.

## Root cause

Two bugs:

**1. Hash encoding corruption (`serialize.ts`)**  
`fillToKiwiPaint` passed the image hash as a hex string (e.g. `'43be04bb...'`) directly into a `byte[]` Kiwi field. `TypedArray.prototype.set(string)` converts each character via `ToUint8()` — hex letter digits like `'b'`, `'e'`, `'f'` become `NaN → 0`. So `'43be04bb...'` was written as `[4, 3, 0, 0, 0, 4, 0, 0, ...]`, corrupting the hash. On reload, `imageHashToString` decoded these corrupted bytes into a different hash that didn't match the key in `graph.images`.

**2. ArrayBuffer detached on save (`export.ts`)**  
`ByteBuffer.toUint8Array()` returns a **view** (not a copy) of the internal buffer. Transferring `kiwiData.buffer` via `worker.postMessage` transferables detached the ByteBuffer's internal buffer, causing "ArrayBuffer at index N is already detached" on subsequent saves.

## Fix

- `serialize.ts`: convert hex hash string to `Uint8Array` before writing to the `byte[]` Kiwi field
- `codec.ts`: update `Paint.image.hash` type to `string | Uint8Array`
- `export.ts`: remove transferables from `postMessage` — data is copied via structured clone instead (safe, since the Worker handles compression, not zero-copy transfer)

## Checklist
- [ ] Tests added or updated


dannote:
Thanks for the detailed report and fix 💛 I pushed a small regression test and shared hex helper on top, and this is ready to go in.

PR #268, opened 2026-05-22
Refactor tooltips and improve internationalization support

SantosAlarcon:
I have added tooltips in some controls and in the AppSelect component, and added some string keys for internationalization.

---

**Checklist:**
- [ ] Tests added or updated


dannote:
Could you please run the formatting?

dannote (approved):
Thanks for this! Pushed a small follow-up to keep the select portal outside the tooltip trigger, make unlabeled tips safe, and keep menu labels reactive.

PR #270, opened 2026-05-24
Fix .fig roundtrip: missing nodes, schema mismatch, GUID collisions

joeycumines:
## Why

Exporting a .fig file from OpenPencil and reimporting it lost nodes. The same exported files also wouldn't open in Figma, including gold-preview.fig which now does. I observed visual issues with roundtripped files even within OpenPencil itself. I haven't traced each symptom to a specific fix with confidence, but the changes address the underlying causes, and produce .fig files that load and appear correct in both Figma and OpenPencil.

## Root causes

1. **fractionalPosition overflow.** The old `String.fromCharCode(33 + index)` produces char code 127 (DEL) at index 94, then characters outside printable ASCII after that. Figma's own position strings never exceed char code 126. I verified that gold-preview.fig has nodes with 492 direct children. What exactly Figma does with DEL and above in position strings, I haven't verified directly, but files with these characters didn't open. The new telescoping `~`-prefix scheme matches Figma's own encoding.

2. **Schema mismatch on re-export.** When re-exporting an imported .fig file, the old code encoded using OpenPencil's subset kiwi schema. This produces different field IDs than the original embedded schema. I believe this means the encoded data is incompatible with the schema Figma expects, though I haven't verified the exact failure mode. The fix preserves the original schema (`graph.figSchemaDeflated`) and compiles it for both encoding and embedding. Likely needs further thought / work / analysis.

3. **GUID collisions between variables and canvas entries.** Variable GUIDs were assigned before canvas entries were built, so monotonically-generated variable GUIDs could collide with `source.id`-derived canvas GUIDs. I haven't verified this actually triggered in practice, but it's clearly possible given the allocation order. The fix reverses the order and scans all imported source.id GUIDs to advance the counter past them.

4. **Child sort ignoring parentIndex.position for auto-layout.** Auto-layout children were sorted only by transform position, which may not reflect the original tree order for imported nodes (e.g. after layout recomputation). Now sorts by `parentIndex.position` first, transform as tiebreaker. I haven't verified a specific visual symptom from this, but wrong layer order seems like a plausible consequence.

5. **rawNodeFields overwriting explicit serialization.** `Object.assign` blindly merged all raw Figma fields into the NodeChange, overwriting fields that were already set correctly by explicit serialization. The fix materializes with paint variables and variable maps, applies a blocklist for structurally dangerous fields, and skips any key already present on `nc`.

## Smaller fixes

- **strokeWeight defaulting**: stopped forcing `strokeWeight=1` on nodes without strokes. This was overwriting the correct raw value for strokeless nodes like text.
- **colorVar.assetRef to guid**: replaced the most-common-color heuristic with direct `key@version` to guid lookup for paint variable references. The old heuristic was guessing colors from usage counts, which seems fragile.
- **corner radii**: always emit per-corner values when present, not just when `independentCorners=true`. Figma can have per-corner values without the flag set.
- **textAlignVertical**: use the actual value instead of hardcoded `'TOP'`.
- **textAutoResize**: preserve original value for imported nodes instead of forcing `'HEIGHT'`.
- **stackReverseZIndex**: roundtrip the field in both import and export.
- **variable key/version**: import from `nc.key`/`nc.version`, re-export on VARIABLE NodeChanges so assetRef resolution works.
- **blendMode on effects**: emit explicitly.
- **textCase**: emit when non-default.

## Test infrastructure

Exhaustive roundtrip test comparing G0 (original file) to G1 (import then export) to G2 (import G1 then export again) across schema bytes, node count, tree paths, node types, scene props, and raw node fields. Three remaining failures are pre-existing known bugs tracked as TODOs: corner radius 999 sentinel lost on scene import (40 nodes), componentPropDefs verifier logic gap (9 instances), derivedTextData baseline precision differences (14 instances, font metrics).

Even if you change this, please keep the strategy of exhaustively validating every property on every node, with as-strict-as-possible heuristics where values aren't 1-1. Roundtripability is probably the most likely thing to make or break the app, IMO.

Some tests fail - some because of actual issues, some likely because of heuristics that need refinement.

## File reorganization

`font-digests.ts`, `font-features.ts`, `font-variations.ts` moved into `font/` subfolder with barrel index, following the project convention for multi-file domains.


dannote:
Awesome work — the roundtrip strategy and schema preservation make sense. Before merging, I want to look into the unknown/missing Figma schema fields first and see if we can add them explicitly to our bundled schema instead of relying only on preserved embedded schemas.

joeycumines:
Sure @dannote , I'll leave it to you 🫡 

Re: the schema strategy, I've a not-quite-suspicion (sinking feeling perhaps, lol) it'll come down to a choice between two less-than-ideal options (strict compliance vs "calculated risk"), since:

1. Safest approach = strict validation of the incoming schema, internal versions included
2. Strictly-compliant well-formed output, using a schema known ahead of time on save
3. There's a time lag between the absolute best case between unknown format -> support, and that's *after* it is known to be broken

On one hand, it is reasonable to assume Figma won't or will very rarely drop support for old schema versions, so it is likely that the core functionality will be reliably supported. On the other, it is a complex domain-specific format. I don't fully understand the role of the schema, but my assumption so far has been it is intended in part to support interacting (to at least some extent) with data that isn't "fully understood". I didnt spend too much time on it, but I'll note I wasn't able to confirm that hypothesis, whether or not that extends to making changes in that scenario, or if it is even wise to attempt.

PR #274, opened 2026-05-25
doc(local-app-build): update CONTRIBUTING.md for build w/o updater key

joeycumines:
Per title - the current updater configuration means it'll error due to the missing key, otherwise.


dannote:
Thanks for this — I highly doubt any human or agent reads that line until the build explodes 😄, but this is a useful fix.

PR #275, opened 2026-05-26
fix(cli): support documented open-pencil command

austinm911:
## Summary

Fixes two CLI package metadata issues:

- resolves Bun source execution for CLI-internal `#cli/*` package imports by mapping them to TypeScript source files explicitly
- exposes the documented `open-pencil` binary name in addition to the existing `openpencil` alias

## Why

The CLI docs and help text refer to the command as `open-pencil`, but the published package only declares `openpencil` as its bin. Also, when running the CLI source with Bun, internal imports such as `#cli/format` can fail because the package import map points to `./src/*` without the `.ts` extension.

## Verification

- `bun packages/cli/src/index.ts --help`
- `bun --filter @open-pencil/core build`
- `bun --filter @open-pencil/cli build`
- `node packages/cli/bin/openpencil.js --help`
- `bunx oxfmt --check packages/cli/package.json`
- `git diff --check`

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -2,29 +2,30 @@
   "name": "@open-pencil/cli",
   "version": "0.12.2",
   "license": "MIT",
-  "type": "module",
-  "imports": {
-    "#cli/*": "./src/*"
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/open-pencil/open-pencil.git",
+    "directory": "packages/cli"
   },
   "bin": {
+    "open-pencil": "./bin/openpencil.js",
     "openpencil": "./bin/openpencil.js"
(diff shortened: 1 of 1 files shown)


dannote:
Thanks for taking a look. I’m going to pass on this PR as-is: `openpencil` has been the released CLI binary since the 0.4.x releases, so the docs are the stale part here rather than the package metadata. I also can’t reproduce the Bun source-import issue on current master (`bun packages/cli/src/index.ts --help` works), and the import-map change doesn’t make Node source execution work either. I’ll update the docs to use the released `openpencil` command instead.

dannote:
Closing in favor of a focused docs correction for the existing released CLI name.

PR #280, opened 2026-05-29
fix(layout): preserve instance child import bounds

im0505:
## Summary

  Fixes imported Figma instance content changing size during OpenPencil layout recomputation.

  When an imported instance is inside an auto-layout parent, `applyYogaLayout()` updated the instance node itself and then recursively applied Yoga layout to the instance's internal children. Those children
  already preserve Figma-authored geometry from the import path, so recomputing them could overwrite fixed imported bounds.

  This preserves instance internals after positioning the instance itself.

  ## Reproduction

  Synthetic repro file: [open-pencil-instance-height-repro.fig.zip](https://github.com/user-attachments/files/28373896/open-pencil-instance-height-repro.fig.zip)

  The file contains:
  - an auto-layout parent frame
  - an instance child
  - a fixed-size auto-layout child inside the instance source
  - large vertical padding metadata on that internal child

  On `master`, importing the file and running layout changes the internal child height from `80` to `3330`. With this patch, the child remains `80`.

  ## Root Cause

  In the attached synthetic repro, a fixed imported child inside an `INSTANCE` imports with raw dimensions of `80 x 80`, but after `computeAllLayouts()` its height could be recomputed from preserved
  vertical padding metadata instead of staying at the imported size.

  The problematic path was:

  - `computeAllLayouts()`
  - `applyYogaLayout()`
  - recursive layout application into `INSTANCE` children

  ## Changes

  - Stop `applyYogaLayout()` from recursing into `INSTANCE` internals after updating the instance node.
  - Add regression coverage for fixed auto-layout children inside instances.
  - Update stale `.fig` roundtrip export size expectations to match current master exporter output.
  - Use the public `#core/kiwi/schema-runtime` entrypoint in schema coverage tests.

  ## Validation

  - `bun run check` passed
  - `bun run format` passed
  - `bun run test:dupes` passed
  - `BUN_HEAVY_TESTS=false bun test tests/engine/` passed
  - Targeted layout/import checks passed
  - Full `bun run test` was run, but current `origin/master` also fails on unrelated E2E baseline issues


dannote (approved):
Thanks for tracking this down — the imported-instance repro is a good catch. I pushed a small follow-up to keep the freeze scoped to
 Figma-imported instances, and the fixture exports still compare cleanly.

PR #281, opened 2026-05-29
fix(core): correct boolean op mapping, rotated transform, and fillGeometry fallback

coder-zkl1988:
## Summary

- Fixes .fig import boolean operation `XOR` → `EXCLUDE` mapping (kiwi enum value 3 was treated as UNION)
- Computes correct `(x,y)` for rotated nodes so `translate+rotate` matches the Figma transform matrix
- Prefers Figma's pre-computed `fillGeometry` over recomputing from children via `Path.op()`, which fails on self-intersecting paths (e.g. star polygons in rating stars)
- Checks `Path.op()` return value and cleans up on failure
- Detects VECTOR nodes with `cornerRadius` + simple 4-vertex straight-segment network and renders as rounded rects

## Test Plan

- [x] Import demo.fig with mixed boolean operations (UNION, SUBTRACT, EXCLUDE/XOR)
- [x] Verify WeChat/Alipay payment icons render with correct cutout shapes (not solid rectangles)
- [x] Verify rotated elements (e.g. divider lines) position correctly and are visible
- [x] Verify star-shaped SUBTRACT boolean ops (rating stars) render with proper cutouts
- [x] Run `bun run lint` — 0 warnings, 0 errors
- [ ] Run `bun test` Playwright E2E visual regression (requires test environment)

### DIFF
--- a/packages/core/src/canvas/boolean.ts
+++ b/packages/core/src/canvas/boolean.ts
@@ -251,6 +251,13 @@ export function makeBooleanOperationPath(
   node: SceneNode,
   graph: SceneGraph
 ): Path | null {
+  const fg = r.getFillGeometry(node)
+  if (fg) {
+    const result = new r.ck.Path()
+    for (const p of fg) result.addPath(p)
+    return result
+  }
+
   const childPaths: Path[] = []
   for (const childId of node.childIds) {
     const child = graph.getNode(childId)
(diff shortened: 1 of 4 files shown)


dannote:
Thanks for digging into this — the XOR mapping and imported geometry fallback were the right direction. I reimplemented the safe parts separately in 6ee0e5f9 with focused regression coverage, while avoiding the stale-geometry and cleanup risks from this branch. Closing this PR in favor of that smaller fix.

PR #282, opened 2026-05-29
Fix MCP server hang after first call

dmitriyi-ai:
## Problem

The MCP server answered only the first tool call per session; every later call timed out (~4 min) with no response, progress, or error. Once `get_page_tree` returned an oversized result, the server stopped responding to everything.

## Root causes & fixes

**1. SSE response mode wedged keep-alive proxies.** The transport returned `text/event-stream` with `Connection: keep-alive`. Proxies like `mcp-remote` (used by Claude Desktop) read the first response, then wait on the half-open stream, so every later request times out. → `enableJsonResponse: true` returns plain JSON that closes cleanly per request.

**2. No server-side response-size guard.** A >1MB body was emitted and rejected client-side, leaving the stream stuck. → `ok()` now measures the serialized result and returns a structured error with a hint instead of emitting an oversized body. Same guard for base64 image results.

**3. `get_page_tree` had no way to bound output.** → added `depth`, `root_id`, `node_types` params (backward compatible — no args = previous behavior).

**4. Zombie app WebSocket caused silent waits.** The browser socket can die without a clean close yet linger in `readyState OPEN`; sends succeed silently and the RPC waits out the full timeout. → 5s ping/pong heartbeat reaps dead sockets; `ws.send` failures reject immediately; single-shot settle prevents double-resolve/leaked timers; RPC timeout 30s→20s.

## Verification

Driven through `mcp-remote` (the Claude Desktop path): 5 sequential tool calls over 42s across multiple heartbeat cycles — each responded in 0.02–0.07s, connection healthy throughout. No hang. Typecheck clean, unit tests pass.

### DIFF
--- a/packages/core/src/tools/read.ts
+++ b/packages/core/src/tools/read.ts
@@ -26,31 +26,66 @@ interface TreeEntry {
   children?: TreeEntry[]
 }
 
-function nodeToTreeEntry(node: FigmaNodeProxy): TreeEntry {
-  const entry: TreeEntry = {
-    id: node.id,
-    type: node.type,
-    name: node.name,
-    w: node.width,
-    h: node.height
-  }
-  if (node.children.length > 0) {
-    entry.children = node.children.map(nodeToTreeEntry)
-  }
-  return entry
-}
-
 export const getPageTree = defineTool({
   name: 'get_page_tree',
   description:
-    'Get the node tree of the current page. Returns lightweight hierarchy: id, type, name, size. Use get_node for full properties of a specific node.',
-  params: {},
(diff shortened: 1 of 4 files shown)


dannote:
Thanks for digging into this. The underlying MCP hardening concerns were valid, but this branch has drifted and isn’t safe to merge as-is. I reworked the safe parts separately in `31a1e171` — response bounds, JSON HTTP responses, browser RPC timeouts, stale WebSocket cleanup, and related hardening — so I’m going to close this PR rather than merge the stale implementation. Appreciate the push on this.

PR #287, opened 2026-06-01
feat(demo): save demo into local storage in development

yoohoo360:
Feat #

- Restore SceneGraph from local cache on demo page startup
- Add Ctrl+Shift+S to persist current graph into cache
- Add Ctrl+Shift+P to print selected nodes or the whole graph
- Enabled only for demo routes with `test` param in development mode

---

**Checklist:**
- [x] Tests added or updated

### DIFF
--- a/packages/core/src/kiwi/index.ts
+++ b/packages/core/src/kiwi/index.ts
@@ -49,3 +49,7 @@ export {
   parseVarint,
   FIG_WIRE_MAGIC
 } from './fig/codec/protocol'
+
+export { deserializeSceneGraph, serializeSceneGraph } from './fig/parse/transfer'
+
+export type { SerializedSceneGraph } from './fig/parse/transfer'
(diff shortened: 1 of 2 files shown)


dannote (changes requested):
I don’t think this should land as-is. The debug workflow is very narrow, but it adds route-specific local-cache behavior directly to EditorView and has a few concrete problems: on /demo?test it uses process.env.NODE_ENV in client code, which is not defined by our Vite config and can throw at runtime; with an empty cache it parses {} and then calls deserializeSceneGraph on it, which will throw because the serialized graph fields are missing; returning false from the keydown handler does not prevent the browser’s Ctrl+Shift+S default, so the save shortcut can still open browser save UI; and dropping the createdInitialTab guard means visiting the demo route can inject demo shapes into an existing active tab. I’d rather not add this as product code without a clearer use case and a safer integration point.

yoohoo360:
The daily workflow of modifying demo code or manually drawing shapes repeatedly is extremely inefficient, and I'm not really willing to touch the demo code, so I ended up writing a piece of code that makes the development and debugging process a bit more efficient.

PR #288, opened 2026-06-01
fix(vue): fix node renaming when the input position changes

yoohoo360:
Fixes #

> fix input value reset when cursor position changes

---

**Checklist:**
- [x] Tests added or updated

## Summary by CodeRabbit

* **Refactor**
  * Improved inline rename workflow: the rename input now reflects live editing state, and commit/cancel reliably clear the editor state and handlers for more consistent behavior.
  * Minor input behavior adjusted to rely on blur for commits, simplifying key interactions during rename.


dannote:
Thanks for the fix. The Layers panel has since been rebuilt, and I confirmed renaming now stays stable on current master, so this is superseded.

PR #295, opened 2026-06-02
fix(canvas): render CJK text via outlines

DENGGL2:
- Bundle Noto Sans SC as a deterministic CJK fallback

- Load the bundled fallback before marking renderer fonts ready

- Route CJK text through the existing outline text renderer when paragraph shaping cannot provide glyphs

Fixes #

---

**Checklist:**
- [ ] Tests added or updated

### DIFF
--- a/packages/core/src/canvas/renderer/fonts.ts
+++ b/packages/core/src/canvas/renderer/fonts.ts
@@ -45,6 +45,9 @@ export async function loadFonts(
     r.fontMgr = r.ck.FontMgr.FromData(fontData) ?? null
   }
 
+  await fontManager.ensureBundledCJKFallback()
+  if (r.isDestroyed()) return
+
   r.fontsLoaded = true
   r.invalidateAllPictures()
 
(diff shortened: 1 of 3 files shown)


dannote (changes requested):
Thanks for looking into CJK rendering. I think the direction needs a bit more design before merging: we already have a CJK fallback path in `FontManager` / paragraph rendering, so I’d rather extend that mechanism instead of adding a separate outline-only path. The current approach also makes renderer startup load a 10 MB font even for documents without CJK text, and rendering all CJK text as outlines can bypass paragraph/text features like shaping, truncation, and styled runs. Could you rework this around the existing fallback-family flow and add a targeted visual regression for the CJK case?

dannote:
Thanks for working on this. I’m going to close it in favor of tracking the broader CJK rendering work in #324. The outline-rendering approach is too risky for normal text behavior, and we should fix this through the existing paragraph/fallback font path instead.

PR #296, opened 2026-06-02
fix(ai): apply stock photos only to image-capable nodes

DENGGL2:
- Reject image fills on text and unsupported node types

- Support data URL photo payloads from image providers

- Expand cover-like placeholders to their parent bounds before applying image fills

Fixes #

---

**Checklist:**
- [ ] Tests added or updated

### DIFF
--- a/packages/core/src/tools/modify/paint.ts
+++ b/packages/core/src/tools/modify/paint.ts
@@ -3,6 +3,16 @@ import { BLACK } from '#core/constants'
 import { defineTool } from '#core/tools/schema'
 import type { Matrix } from '#core/types'
 
+function canUseImageFill(type: string): boolean {
+  return (
+    type === 'RECTANGLE' ||
+    type === 'ELLIPSE' ||
+    type === 'FRAME' ||
+    type === 'COMPONENT' ||
+    type === 'INSTANCE'
+  )
+}
+
 export const setFill = defineTool({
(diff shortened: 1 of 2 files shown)


dannote (changes requested):
Thanks for chasing this — I agree there’s a real issue around stock photos being applied to unsuitable targets. I don’t think the name-based resize heuristic is the right fix, though. Expanding nodes named `background`, `cover`, `hero`, `header`, etc. to their parent bounds can unexpectedly mutate user geometry, and it also computes the requested image size before that resize, so small placeholders may still fetch low-res images and then stretch them. I’d prefer keeping this more explicit: reject truly unsupported targets, and let the agent use the existing resize/update tools when it wants a placeholder to cover its parent. The image-capable allowlist also looks too narrow — it rejects valid fillable shapes like rounded rectangles/vectors/polygons/boolean ops. Data URL support seems useful, but the target validation/resize behavior needs a less heuristic pass before merging.

dannote:
Thanks again for identifying the target-validation problem. This branch is now substantially behind the current codebase, and the requested changes were not addressed, so I’m closing it as superseded by #639.

The replacement keeps the useful validation intent, adds current regression coverage, and avoids the name-based resizing, narrow fillable-node allowlist, and custom data URL decoding proposed here.

PR #312, opened 2026-06-04
fix(canvas): lazy load CJK fallback fonts

DENGGL2:
Fixes #

---

**Checklist:**
- [ ] Tests added or updated

Thanks for the review. I updated the approach to follow the existing paragraph/font fallback path instead of rendering CJK text through a separate outline-only path.

Changes in the new branch:
- Removed eager CJK/Arabic fallback loading during renderer startup.
- Added lazy fallback requests when text rendering actually encounters missing CJK/Arabic fallback fonts.
- Trigger the same lazy fallback path while editing text, so manually typed Chinese text repaints once the fallback is ready.
- Extended ensureGraphFonts so AI/automation-created CJK text also loads fallback fonts on demand.
- Updated the CJK fallback regression test to assert fallback fonts are not loaded at startup or for Latin text, and are only requested after CJK text is present.

I manually verified that typing Chinese text in the editor now renders correctly.

Note: this reply was drafted with AI assistance and reviewed before posting.

### DIFF
--- a/packages/core/src/canvas/index.ts
+++ b/packages/core/src/canvas/index.ts
@@ -6,3 +6,4 @@ export {
 } from './boolean'
 export { SkiaRenderer, type RenderOverlays, type RulerTheme } from './renderer'
 export { getAbsolutePositionFull, getAbsoluteRotation, getWorldHandles } from './coordinate'
+export { ensureMissingFallbackFonts } from './text'
--- a/packages/core/src/canvas/renderer.ts
+++ b/packages/core/src/canvas/renderer.ts
@@ -77,6 +77,8 @@ export class SkiaRenderer {
   fontMgr: FontMgr | null = null
   fontProvider: TypefaceFontProvider | null = null
   fontsLoaded = false
+  onFallbackFontsLoaded: (() => void) | null = null
+  pendingFallbackScripts = new Set<'cjk' | 'arabic'>()
   imageCache = new Map<string, CKImage>()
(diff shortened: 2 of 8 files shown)


dannote (changes requested):
Thanks, this is much closer to the direction we want: using the existing paragraph/fallback path and loading fallback fonts only when needed. I’d like to tighten two things before merging.

First, the updated regression test no longer exercises the renderer lazy-load path. It calls `ensureGraphFonts()` directly, so the test would still pass if `renderText()` stopped calling `ensureMissingFallbackFonts()` or if the fallback-load callback stopped repainting the canvas. The previous test at least covered the delayed fallback callback/re-render behavior. Could you add or restore coverage where a CJK text node is rendered while fallback fonts are missing, the fallback promise resolves later, and the renderer repaint callback fires? Ideally it should also assert visible output or a snapshot rather than only font-manager state.

Second, this adds another CJK regex in `src/app/editor/fonts/index.ts`, and it already differs from the broader ranges being discussed for Hangul/CJK coverage. Since fallback detection now exists in multiple layers, please centralize the script detection helper in core and reuse it from both canvas text and app font loading so we don’t keep drifting between slightly different CJK definitions.

dannote:
Thanks for working on this. The font pipeline has since been rebuilt on `master`, including lazy CJK/Arabic fallbacks and repaint coverage, so this PR is now superseded.

PR #316, opened 2026-06-05
fix(canvas): stabilize CJK paragraph rendering

im0505:
## Summary

- Stabilizes CJK paragraph rendering by using a CanvasKit `FontMgr` built from already-loaded font buffers for CJK text.
- Keeps paragraph `FontMgr` instances alive through a bounded renderer cache so repeated renders/moves do not reuse paragraphs backed by deleted CanvasKit objects.
- Adds regression coverage for CJK fallback readiness and distinct CJK glyph rendering across repeated rerenders.

## Reproduction

No `.fig` sample is attached. The original failing document is private, and the minimized regression does not require a file fixture. The added E2E test creates a minimal public CJK text node directly and exercises the failing render path.

Manual reproduction:

1. Open any OpenPencil document.
2. Create a text node with a primary font that does not cover CJK glyphs, such as `Inter`.
3. Enter CJK text such as `환경설정` or `你好世界`.
4. Ensure a CJK fallback font is available.
5. Move the text node repeatedly or otherwise trigger repeated canvas renders.

Before this change, the text could render as `NO GLYPH`, and repeated rendering/moving could crash CanvasKit with `table index is out of bounds`.

After this change, CJK text renders through the paragraph shaper and repeated movement does not crash.

## Implementation notes

This does not add a new network font loading path. The new CJK paragraph path only uses font buffers that are already loaded in `fontManager`.

The `FontMgr` cache is bounded and cleared when fonts reload or the renderer is destroyed. Non-CJK text, including Arabic-only text, continues to use the existing paragraph builder path.

## Testing

- `bun run check`
- `bun test tests/engine/render/canvas/text.test.ts tests/engine/render/canvas/font-readiness.test.ts`
- `bun run test -- tests/e2e/text/cjk-rendering.spec.ts`

**Checklist:**
- [x] Tests added or updated

## Summary by CodeRabbit

* **New Features**
  * Per-renderer paragraph font-manager caching to improve CJK shaping stability and rendering consistency.

* **Bug Fixes**
  * Clear and release paragraph font-manager cache during font reloads and renderer teardown to prevent leaks.
  * Improved font-readiness detection for decomposed Hangul and other complex scripts.

* **Tests**
  * Added visual e2e and unit tests validating CJK glyph distinctness, fallback behavior, and renderer lifecycle.


dannote (changes requested):
Thanks for digging into this — the CanvasKit lifetime issue sounds real, and keeping `FontMgr` instances alive may be the right general direction. I don’t think this is safe as-is, though. The new CJK path switches to `FontMgr.FromData(...)` whenever any loaded font data exists, which means a document with loaded `Inter` but a CJK fallback registered only on the existing font provider now bypasses that provider and builds a paragraph font manager without the fallback. The added `markLoaded("Noto Sans SC", ...)` in the unit test hides that regression; previously `fontProvider.registerFont(...)` plus `setCJKFallbackFamily(...)` was enough. If we use this path, it should only activate when all required primary/style/fallback fonts are available as loaded data, otherwise fall back to the existing provider path.

The new Hangul E2E also doesn’t prove the rendering fix: it sets `Inter` as the CJK fallback even though `Inter` has no Hangul glyphs, then only asserts that the node text is unchanged and no console errors occurred. That would still pass with tofu/blank rendering. Could you update it to load a real Hangul-capable fallback and assert visible output or use a snapshot?

im0505:
Thanks for the review. I updated the PR to address the concerns.

- The `FontMgr.FromData(...)` path now only activates when all required primary/style/fallback font buffers are already loaded. If any required buffer is missing, it falls back to the existing provider-based paragraph path, preserving provider-only CJK fallbacks.
- Added/updated regression coverage for the provider-only fallback path and isolated the global fallback state between tests.
- Updated the E2E coverage to use a CJK glyph rendering test with a local `NotoSansSC-Regular.ttf` fixture. It renders `你`, `好`, `世`, and `界` separately and compares visible canvas pixel signatures, so tofu/NO GLYPH-style repeated glyph output should fail.
- Tightened CanvasKit cleanup around paragraph builder/paragraph creation and clipboard shaping error paths.
- This does not add a new runtime network font loading path.

Validation:
- `bun run check`
- `bun test tests/engine/render/canvas/text.test.ts tests/engine/render/canvas/font-readiness.test.ts tests/engine/render/canvas/renderer-lifecycle.test.ts`
- `bun run test -- tests/e2e/text/cjk-rendering.spec.ts`

CI is green on the latest head (`9a595d24`). Could you take another look?

dannote (dismissed):

IcyVizi:
image
Any plan to fix emoji font rendering?

PR #317, opened 2026-06-05
fix(resize): scale children when resizing GROUP or BOOLEAN_OPERATION

rcoenen:
## Problem

Resizing a **group** (or a **boolean operation**) only moved the bounding box and rescaled the selection outline — the rendered fill stayed at its original size. The blue selection outline and the filled content visibly diverged, both during the drag and after release.

### Before
Group resize — outline scales but fill stays small

## Root causes

Both in the resize preview path:

1. **Stale render caches during drag.** Children were scaled in the preview graph via `updateNodePreview()`, which mutates nodes **without** emitting a `node:updated` event. The renderer's vector/geometry caches (`vectorPathCache`, `fillGeometryCache`, `strokeGeometryCache`) were therefore never invalidated, so fills kept drawing from stale, unscaled cached paths while outlines drew live from the new dimensions.

2. **No repaint on commit.** On mouse-up the final scaled values were written with `updateNode()`, but no repaint was requested — leaving the canvas frozen on the last mismatched drag frame until the next interaction.

## Fix

- After each child preview update, call `renderer.invalidateVectorPath(childId)` so the renderer recompiles paths from the scaled coordinates every frame.
- Call `requestRepaint()` after `commitGroupResize()` so the canvas redraws immediately on release.
- Extend `collectDescendants()` to also handle **`BOOLEAN_OPERATION`** nodes. Boolean results render purely from child geometry with no node-level scale factor (`nodePathTransform` only applies translate/rotate/flip), so scaling the children is the only way to scale the result — the same mechanism as `GROUP`, with no risk of a double-transform.

### After
Group resize — fill and outline scale together

## Testing

- **Group resize** — verified visually (108→111, 130×130): fill and per-child outlines scale in lockstep, during drag and after release.
- **Boolean resize** — verified in-browser via the editor automation API (create two shapes → SUBTRACT → resize): the combined boolean fill scales with the bounding box.

### DIFF
--- a/packages/core/src/editor/bridges/undo.ts
+++ b/packages/core/src/editor/bridges/undo.ts
@@ -10,6 +10,7 @@ export function createUndoBridge(undoActions: UndoActions, selection: SelectionA
     commitMoveWithReparent: undoActions.commitMoveWithReparent,
     commitDuplicateMove: undoActions.commitDuplicateMove,
     commitResize: undoActions.commitResize,
+    commitGroupResize: undoActions.commitGroupResize,
     commitRotation: undoActions.commitRotation,
     commitNodeUpdate: undoActions.commitNodeUpdate,
     undoAction: () => undoActions.undoAction(selection.validateEnteredContainer),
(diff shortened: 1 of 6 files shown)


dannote (approved):

dannote:
Merged, thanks 💛

PR #318, opened 2026-06-05
fix(boolean-ops): make operations dropdown menu appear on screen

rcoenen:
## Problem

The boolean operations dropdown (Union / Subtract / Intersect / Exclude / Flatten) **opened but rendered off-screen** — its menu items were positioned at roughly `x=5, y=-295…-183`, entirely above the top-left of the window — so the menu was effectively invisible and the feature was unusable from the GUI. (Keyboard shortcuts ⌥⇧U/S/I/E/F still worked, which is why it wasn't caught.)

## Root cause

In `BooleanOperationsControl.vue` the trigger button was wrapped in `<Tip>` (a `Tooltip` + `TooltipTrigger as-child`) **between** `<DropdownMenuRoot>` and `<DropdownMenuTrigger as-child>`:

```
<DropdownMenuRoot>
  <Tip>                              <!-- TooltipTrigger as-child -->
    <DropdownMenuTrigger as-child>   <!-- second as-child on the same button -->
      <button>…</button>
    </DropdownMenuTrigger>
  </Tip>
  <DropdownMenuPortal>…</DropdownMenuPortal>
</DropdownMenuRoot>
```

Chaining **two `as-child` triggers** onto the same button means the `DropdownMenuTrigger` never gets a clean ref to the button element, so the dropdown's floating content has no anchor to position against and falls back to an off-screen default.

## Fix

Match the working pattern used elsewhere in the app (e.g. `VariablesDialog.vue`): let `DropdownMenuTrigger` wrap the `<button>` **directly**, with no tooltip trigger in between. The tooltip label is preserved via a native `title` attribute on the button.

```
<DropdownMenuRoot>
  <DropdownMenuTrigger as-child>
    <button :title="commands.booleanOperations">…</button>
  </DropdownMenuTrigger>
  <DropdownMenuPortal>…</DropdownMenuPortal>
</DropdownMenuRoot>
```

## Testing

Verified in-browser: with two shapes selected, clicking the trigger now opens the menu **directly below the button** (items on-screen at `x≈1205, y≈126–238`), and selecting an operation performs the boolean as expected.

### DIFF
--- a/src/components/properties/BooleanOperationsControl.vue
+++ b/src/components/properties/BooleanOperationsControl.vue
@@ -16,7 +16,6 @@ import IconSquaresIntersect from '~icons/lucide/squares-intersect'
 import { editorCommandMetadata, formatShortcut, useEditorCommands, useI18n } from '@open-pencil/vue'
 import type { EditorCommandId } from '@open-pencil/vue'
 
-import Tip from '@/components/ui/Tip.vue'
 import { menuItem, useMenuUI } from '@/components/ui/menu'
 
 const { getCommand, runCommand } = useEditorCommands()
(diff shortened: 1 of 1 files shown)


dannote:
Thanks for tracking this down. I pushed a more general fix in #325 that keeps `<Tip>` working while avoiding nested Reka trigger ownership, so it covers this boolean menu case plus the same issue in selects/popovers. Closing this one as superseded, but the root-cause write-up here was helpful.

PR #320, opened 2026-06-05
fix(kiwi): import Figma groups as GROUP nodes instead of FRAME

rcoenen:
- Map FRAME nodes with `resizeToFit === true` to `GROUP` type on import
- Write `resizeToFit = true` when exporting `GROUP` nodes back to Kiwi format
- Document findings and diagnosis in packages/docs/development/group-to-frame-import-issue.md


dannote (approved):

dannote:
Nice fix, thanks 💛

PR #321, opened 2026-06-05
feat(export): per-node settings, editable scale, JPG/WEBP, zip bundling

rcoenen:
## Why

On the deployed app, the **Export panel was broken and not Figma-parity**:

- The **format dropdown** rendered off-screen (especially with multiple export rows), so you could not switch a row from PNG to JPG/WEBP/SVG/PDF.
- **JPG/WEBP** exports failed with "Nothing to export" and produced no file (CanvasKit `encodeToBytes` returns `null` for those formats in this build).
- Export settings were **global/transient** — every selection looked like it already had a 1× PNG row, and settings were not remembered per layer or saved with the document.
- **Scale** was preset-only; Figma lets you type arbitrary multipliers.

This PR fixes those bugs and brings export behavior in line with Figma: same panel flow, per-layer settings, persistence, and sensible downloads.

### Per-node export settings (Figma parity)

- Export settings are **empty by default** — header + `+` only until the user adds a row.
- Settings are stored on each **`SceneNode`** (`exportSettings: ExportSetting[]`), applied across multi-select, and target the **current page** when nothing is selected.
- **Persist** with the document via `open-pencil` `pluginData` in `.fig` (lossless, including `webp`).
- On import, read app `pluginData` first; otherwise map native Figma `exportSettings` (PNG/JPEG/SVG/PDF, content scale) without overwriting raw native fields on re-export.
- All add/edit/remove operations are **undoable**.

**Core:** `packages/core/src/scene-graph/types.ts`, `plugin-data.ts`, `convert.ts`, `export-node.ts`  
**Vue SDK:** `packages/vue/src/document/export/{use,helpers}.ts`  
**UI:** `ExportSection.vue`

### Editable export scale (Figma parity)

- New **`ExportScaleInput.vue`**: type any multiplier (e.g. `9x`, `1.5x`) or pick presets from a chevron menu.
- Clamped to **0.01×–1024×** in `clampExportScale` (defends renderer from huge canvases); invalid input reverts.

### Format dropdown (`AppSelect`)

- `AppSelect` no longer wraps the trigger in `<Tip>` (that broke reka-ui popper anchoring). Tooltips use the native **`title`** attribute on `SelectTrigger` instead.
- Dropdown menus anchor correctly across the inspector (export format, typography, effects, layout, fills, etc.).

### JPG/WEBP export

- **`encodeRasterFallback()`** on the renderer: when CanvasKit cannot encode JPEG/WEBP, encode pixels through an `HTMLCanvasElement`.
- Wired from `packages/core/src/io/formats/raster/render.ts`.
- **Known limitation:** the fallback uses the browser `<canvas>`, so it only applies in-browser. Headless/CLI export of JPG/WEBP still returns `null` — `PNG`/`SVG` are unaffected. A Node-side JPEG/WEBP encoder is left as follow-up; this PR fixes the reported in-app bug.

### Download bundling

- Export collects all rows for the active target(s) and calls **`exportTargets()`**.
- **One file** → direct download (unchanged UX).
- **Multiple files** (several rows, formats, or layers) → **one `.zip`** named after the layer; duplicate filenames disambiguated.

### Scale stepping when adding rows

- Each new export row defaults to **2× the previous scale** (1× → 2× → 4× …), keeping the last row's format.

## Tests

- **Engine:** `tests/engine/io/fig/roundtrip/export-settings.test.ts` — pluginData round-trip (`webp`), native Figma PNG scale mapping, empty default serialization.
- **E2E:** `tests/e2e/export/basic.spec.ts` — empty default, add/remove rows, format/scale edits, multi-select + mixed indicator, undo, multi-row format dropdown, zip bundling, single-file direct download.
- `bun run check` and CI unit tests (`BUN_HEAVY_TESTS=false`) green; verified live (JPG/WEBP files, zip contents, dropdown on-screen).

## Commits (newest first)

1. `fix(export): working format dropdown, JPG/WEBP export, and zip bundling`
2. `feat(export): store export settings per node`
3. `feat(export): editable custom scale with 1024x cap`
4. `fix(ui): anchor AppSelect dropdown to its trigger` — superseded in final `AppSelect` by removing `<Tip>` + native `title`; kept in history for the initial popper investigation.


dannote:
Thanks for the detailed export work. I split out the tooltip/select anchoring issue into #325 so `AppSelect` can keep using `<Tip>` safely instead of switching to native `title`. I’ll review the remaining export behavior separately from that UI root cause.

dannote (changes requested):
Thanks for the big export pass — there’s useful work here, but I found a few correctness issues that should be fixed before merge.

- `packages/core/src/kiwi/fig/node-change/plugin-data.ts:42` returns early for empty `exportSettings`, while `packages/core/src/scene-graph/source-metadata.ts:7` does not clear raw native `exportSettings`. Import a Figma node with native export settings, remove all rows, save/reopen: the raw/native settings can come back because no plugin override is written and import falls back to `nc.exportSettings` at `plugin-data.ts:140`.
- `packages/core/src/kiwi/fig/node-change/plugin-data.ts:104` and `:130` accept any finite scale from plugin/native data. The UI clamps edits, but malformed/imported `.fig` data can still persist huge scales and later allocate via export. Please clamp or reject at import/file-format boundaries too.
- Multi-select export state can diverge: `packages/vue/src/document/export/helpers.ts:66` renders rows from only the first target, `src/components/properties/ExportSection.vue:163` hides the button when that first target has no rows, but `doExport()` exports every selected node’s actual rows at `ExportSection.vue:50-58`. That means hidden settings may export, or settings on non-first selected nodes may be impossible to export from the UI.
- ZIP entries use raw layer/page names: `src/app/document/export/files.ts:34`/`:44` writes `file.fileName` directly into the zip. Names with `/`, `\`, `..`, or control characters should be sanitized to flat safe filenames before zipping.
- The JPEG/WEBP fallback at `packages/core/src/canvas/renderer.ts:644-645` does not verify `toDataURL()` honored the requested MIME. Browsers may return PNG for unsupported encoders, which would save PNG bytes with a `.webp` extension.

Also, after #325, the `AppSelect`/native `title` part is stale — please rebase and keep the shared `<Tip>` path instead of carrying that workaround here.

rcoenen:
Thanks for the detailed review 🙏 Rebased on master (kept the shared `<Tip>` per #325 — dropped the native-`title` change) and addressed all five:

- **Native export settings resurrecting** — `clearEditedSourceMetadata` now drops the raw native `exportSettings` when a node's settings are edited/cleared, so they don't come back via the import fallback on reopen.
- **Unclamped scale at the file-format boundary** — moved the bounds into `core/scene-graph/export-scale.ts` (single source of truth, reused by the UI) and clamp both plugin and native `CONTENT_SCALE` on import.
- **Multi-select divergence** — export now follows the panel (WYSIWYG): it exports the shown `activeSettings` for every target, and the "mixed" note shows whenever targets differ — so no hidden rows export.
- **Zip entry names** — sanitized (strip separators, `..`, and control chars) before zipping.
- **JPEG/WEBP fallback** — verify `toDataURL()` returned the requested MIME; reject a silent PNG so we never write PNG bytes under a `.jpg`/`.webp`.

Added regression tests for the native-clear and import-clamp. CI green. 🙇

dannote (approved):

PR #326, opened 2026-06-06
fix(desktop): drag-and-drop + vectorize CORS in the desktop app

rcoenen:
Two desktop-only fixes found while testing the bundled app.

> [!NOTE]
> **Stacked on #322** (`feat/image-vectorization`). The CORS fix edits that PR's
> `providers.ts`, so this should merge after #322. Until then the diff below also
> shows #322's commits.

### 1. HTML5 drag-and-drop was dead in the desktop app
Tauri v2 enables the OS-level webview drag-drop handler by default, which swallows
file drops before the DOM `drop`/`dragover` events fire — so dropping an image onto
the canvas did nothing. Set `dragDropEnabled: false` so the existing `useCanvasDrop`
handler receives the drop. (Pre-existing bug, affects every desktop build.)

### 2. Vectorize was CORS-blocked in the desktop app
Recraft/fal only whitelist the web dev origin (`http://localhost:1420`); the desktop
webview origin (`tauri://localhost`) is rejected, so vectorize failed with a
network/CORS `TypeError`. Route provider requests through the Tauri HTTP plugin when
running under Tauri (issued from Rust, no browser Origin); fall back to platform
`fetch` in the web build. Adds `@tauri-apps/plugin-http` + the Rust crate, plugin
registration, and a scoped `http:default` capability for the provider hosts.

Both verified live in the bundled macOS app.


dannote:
Thanks, Rob! The drag-and-drop fix is on `master`, and vectorization now uses the secure desktop HTTP bridge. Your contribution credit is preserved. This PR is superseded, so I’m closing it.

PR #333, opened 2026-06-07
fix(export): prevent GUID collisions and file corruption on .fig round-trip

### LINKED ISSUE #349: `.fig` writer silently destroys existing nodes when new nodes are added (sessionID:1 GUID collision)
## Summary

When you add new nodes to a `.fig` and save, then add more nodes and save again, **the second batch silently destroys the first** (and any other `sessionID:1` content). New nodes are minted at `{ sessionID: 1, localID: <small counter> }`, colliding 1:1 with existing `sessionID:1` nodes; the Kiwi reader resolves by GUID as last-write-wins, so each collision overwrites a real node. **Adding N new nodes destroys ~N existing nodes** — silent data loss on basic incremental editing.

Two common triggers:
- **Incremental programmatic edits** (CLI `eval`, scripted builds): nodes created in one write become `sessionID:1`; the next write's new nodes reuse the same low localIDs and clobber them. Verified with the minimal repro below — **no Figma import required**.
- **Documents imported/pasted from Figma**, where *all* content is already `sessionID:1`, so the very first added node destroys existing content. Likely the underlying cause of #102.

## Steps to reproduce (minimal, CLI v0.13.2 — no Figma import needed)

Start from any `.fig` with one empty page. `CLI=…/@open-pencil/cli/dist/index.mjs`

**Write A — add 5 frames:**
```bash
bun "$CLI" eval doc.fig -w -c '
const pg = figma.currentPage;
for (let i=0;i<5;i++){ const f=figma.createFrame(); f.name="A"+i; f.resize(80,80);
  f.fills=[{type:"SOLID",color:{r:1,g:0,b:0}}]; pg.appendChild(f); f.x=10+i*90; f.y=10; }'
# reload: these 5 frames now carry source.ids 1:82 .. 1:86  (sessionID 1)
```

**Write B — add 5 more frames (separate write):**
```bash
bun "$CLI" eval doc.fig -w -c '
const pg = figma.currentPage;
for (let i=0;i<5;i++){ const f=figma.createFrame(); f.name="B"+i; f.resize(80,80);
  f.fills=[{type:"SOLID",color:{r:.5,g:.5,b:.5}}]; pg.appendChild(f); f.x=10+i*90; f.y=110; }'
```

**Inspect:**
```bash
bun "$CLI" eval doc.fig -c 'console.log(figma.currentPage.children.map(n=>n.name).join(","))'
# → "B0,B1,B2,B3,B4"   (A0..A4 are GONE — expected 10 children, got 5)
bun "$CLI" export doc.fig --page "<name>" -o out.png   # PNG shows only the 5 grey B frames
```

**Observed:** 5 children, not 10. The Write-B frames were minted `1:82..1:86`, collided with the Write-A frames, and replaced them. (Don't judge by top-level counts alone in nested cases — a destroyed node can be a deep descendant; scan all emitted GUIDs and/or export to PNG.)

## Expected vs actual
- **Expected:** adding new nodes never alters or deletes existing nodes; total node count increases by the number added.
- **Actual:** new `sessionID:1` nodes collide with existing `sessionID:1` GUIDs; last-write-wins destroys existing nodes; content is silently lost on save.

## Root cause

Two places in `packages/core/src` disagree about which session the collision-prevention covers.

**1. New nodes are minted in `sessionID: 1`** — `packages/core/src/kiwi/fig/node-change/export-node.ts` (`getOrCreateNodeGuid`):
```ts
const guid = importedGuid ?? { sessionID: 1, localID: localIdCounter.value++ }
```

**2. But the high-water-mark scan that advances `localIdCounter` to avoid collisions only looks at `sessionID === 0`** — `packages/core/src/io/formats/fig/export.ts`:
```ts
// Scan ALL imported source.ids to find max sessionID:0 localID,
// preventing collisions between variable GUIDs and any imported node GUID.
let maxLocalId0 = localIdCounter.value - 1
for (const node of graph.nodes.values()) {
  if (node.source.id) {
    const guid = stringToGuid(node.source.id)
    if (guid.sessionID === 0 && guid.localID > maxLocalId0) {   // ← ignores sessionID:1
      maxLocalId0 = guid.localID
    }
  }
}
localIdCounter.value = Math.max(localIdCounter.value, maxLocalId0 + 1)
```

The comment says "Scan **ALL** imported source.ids," but the `guid.sessionID === 0` filter excludes exactly the session (`1`) that new nodes are minted into. So `localIdCounter` is never advanced past existing `sessionID:1` content, and new nodes are minted at colliding localIDs (`1:82, 1:83, …`). The shared `localIdCounter` also mints `sessionID:0` canvas/variable GUIDs, so its localID space spans both sessions — which is why the scan needs to consider the full localID range, not one session.

## Proposed fix

Make the high-water-mark scan cover the session that new node GUIDs actually use — track the max `localID` across **all** source.ids (matching the comment's stated intent), not only `sessionID === 0`:
```ts
let maxLocalId = localIdCounter.value - 1
for (const node of graph.nodes.values()) {
  if (node.source.id) {
    const guid = stringToGuid(node.source.id)
    if (guid.localID > maxLocalId) maxLocalId = guid.localID   // drop the `sessionID === 0` filter
  }
}
localIdCounter.value = Math.max(localIdCounter.value, maxLocalId + 1)
```
(Because `localIdCounter` feeds both `sessionID:0` canvas/variable GUIDs and `sessionID:1` node GUIDs, the safe high-water mark is the max localID over all GUIDs regardless of session. A cleaner long-term fix would be separate per-session counters.)

I verified a userland version of this (assigning each new node a `source.id` in a fresh, higher session before saving): the same 5+5 sequence then keeps all 10 nodes, 0 duplicate GUIDs.

I'm happy to open a PR with this change plus a regression test (import/create a `sessionID:1` fixture → add a node → export → assert zero duplicate GUIDs, all original IDs present, total count = original + 1) if that's helpful.

## Version
openpencil / @open-pencil/core / @open-pencil/cli **0.13.2**.

Possibly the same root cause as #102.

joeycumines:
## Why

I observed data corruption occurring over time during continuous agentic workload iterations. This changeset appears to have resolved the issue.

Exporting a `.fig` file could silently lose nodes. When two nodes shared the same imported `source.id` (which happens naturally with component instance children), the GUID assignment reused the same value. This caused a last-write-wins collision on reimport. Additionally, cloning a node shared the original's `source.fig` object reference, meaning mutations on a clone corrupted the original data structure.

Fixes #349

## What

* **GUID collision prevention:** Tracks every assigned GUID in a `Set<string>` for O(1) duplicate detection. It scans all imported `source.id`s (across both sessionID 0 and 1) before any new GUID is claimed. When a `source.id` GUID collides with an already-assigned value, it falls back to a counter-based GUID instead.
* **cloneTree deep copy:** Uses `structuredClone(source.fig)` instead of sharing the object reference. Mutations on a clone no longer affect the original.
* **Decompression error handling:** Removes the `try/catch` fallback in `parseFigKiwiContainer` that silently treated failed zlib inflation as raw bytes. Corrupt data now explicitly throws. Removed the unused `decompressFigKiwiData` synchronous wrapper.
* **Zstd compression on export:** Modifies `buildFigKiwi` to use `Bun.zstdCompressSync` when running under Bun, matching the zstd decompression path already present on import. This yields an approximate 17% reduction in export size on `gold-preview.fig`.
* **Dirty-state fix:** Moves `setSavedVersion` to run after `requestRender` in `read.ts`, capturing the post-bump version number so the document does not falsely appear unsaved immediately after a reload.

## Scope

| Area | Change |
| --- | --- |
| `packages/core/src/io/formats/fig/export.ts` | GUID collision tracking, sessionID 0+1 scan |
| `packages/core/src/kiwi/fig/node-change/export-node.ts` | Collision detection in `getOrCreateNodeGuid` |
| `packages/core/src/kiwi/fig/node-change/serialize.ts` | Thread `assignedGuidValues` through |
| `packages/core/src/scene-graph/index.ts` | Deep clone `source.fig` in `cloneTree` |
| `packages/core/src/kiwi/fig/container/kiwi.ts` | Zstd compress, remove dead sync wrapper |
| `packages/core/src/kiwi/fig/parse/core.ts` | Remove silent decompression fallback |
| `src/app/document/io/read.ts` | Fix `setSavedVersion` ordering |
| `packages/core/src/index.ts` | Remove dead `decompressFigKiwiData` re-export |

## Summary by CodeRabbit

* **New Features**
  * Prevents duplicate ID conflicts during Figma export/import so separate nodes are preserved on roundtrip
  * Adds boolean shape operation support in exports

* **Improvements**
  * More robust FIG decompression with safer detection and fallbacks
  * Improved FIG compression, reducing export sizes (~17% smaller)

* **Tests**
  * Added GUID-collision and parse-failure suites; extended several test timeouts for stability


dannote (dismissed):

joeycumines:
@coderabbitai re-review requested — all previous findings (formatter fix in df9a0520, type guard in 51ce5b22) have been addressed in subsequent commits. Please re-review the latest commit.

joeycumines:
@coderabbitai re-review requested — commit f6ee766a addresses the docGuid reservation finding and adds a session-0 regression test. The async keyword was also removed from the component-metadata test. Please re-review.

joeycumines:
@coderabbitai re-review requested — commit 3f5eee44 adds assignedGuidValues collision guard in buildCanvasEntries() for canvas GUIDs, mirroring the node GUID logic. Please re-review.

joeycumines:
@coderabbitai full review

joeycumines:
## CodeRabbit Review Response — PR #333

Addressing all CHANGES_REQUESTED items from the latest CodeRabbit reviews:

### ✅ Fixed (commit `29b9cd2f`)

1. **`tests/engine/io/fig/export/parse-failures.test.ts` — Little-endian writes**: Replaced all host-endian `Uint32Array` writes with `DataView.setUint32(offset, value, true)` for platform-independent fig-kiwi container assembly.

2. **`tests/engine/scene-graph/clone/tree.test.ts` — Misleading test title**: Renamed from "clone clears source.id from the original" → "clone clears source.id from the clone" to match the assertions.

### ✅ Already addressed (verified in current code)

3. **`packages/core/src/io/formats/fig/export.ts` — Document GUID reservation**: Already implemented at lines 341-343: `assignedGuidValues.add(\`${docGuid.sessionID}:${docGuid.localID}\`)`. The session-0 collision test at `guid-collision.test.ts:101` confirms this works.

4. **`packages/core/src/io/formats/fig/export.ts` — Canvas GUID collision guard**: Already implemented at lines 247-256: the IIFE checks `assignedGuidValues.has(key)` and falls back to a fresh GUID on collision.

5. **`guid-collision.test.ts` — Session-0 namespace test**: Already exists (test "nodes with session-0 source.id do not collide with document GUID"). All 4 collision tests pass.

6. **`component-metadata.test.ts` — Unnecessary async**: The test callback was already non-async in the current code.

joeycumines:
@coderabbitai full review — all CHANGES_REQUESTED items have been addressed in commit 29b9cd2f. Please re-review.

dannote (approved):

PR #359, opened 2026-06-23
Use a compact, tool-forcing system prompt for custom-endpoint providers

seanshin:
## Problem

The design system prompt (`system-prompt.md`) is rich with inline JSX examples and a multi-phase "plan in text first" workflow. Frontier models follow it well, but **smaller / self-hosted models** reached through the **OpenAI-compatible** and **Anthropic-compatible** providers (Ollama, vLLM, LM Studio, ...) tend to **imitate the JSX examples as plain text** instead of emitting tool calls. The result: the chat narrates a whole design, the tool log shows nothing, and **the canvas stays empty**.

I hit this running OpenPencil against a local Ollama (`qwen2.5:7b`/`14b`) endpoint via the OpenAI-compatible provider. Reproduced the "narrate-instead-of-call" behavior consistently with the full prompt; a short, tool-forcing prompt fixed it.

## Change

Add a compact, tool-forcing system prompt (`system-prompt-compact.md`) and use it **only** for the `openai-compatible` and `anthropic-compatible` providers — the "bring-your-own-endpoint" providers that are typically pointed at local/smaller models. It keeps the props/layout rules but:

- forbids narrating JSX as text — forces immediate `render` tool calls
- builds the whole screen inline in `render` and avoids ID-dependent modify tools (`set_layout`/`batch_update`/...) that small models can't reliably target (they hallucinate node IDs)
- avoids `<Icon>` and non-existent components (`<Button>`, `<Input>`, ...) whose guessed names throw `Icon "x" not found` / `X is not defined` and fail the whole render — uses primitive recipes instead
- requires a concrete root size and uses solid-color placeholders instead of the `stock_photo` tool (which needs an API key that may not be configured)

**Behavior for the hosted providers (OpenRouter / Anthropic / OpenAI / Google / ...) is unchanged** — they keep the full prompt.

## Why route by provider (and not a toggle)

The compatible providers exist specifically for custom endpoints, which are overwhelmingly local/smaller models, so this is a sensible default with no UI/i18n surface. If maintainers would rather make it an explicit opt-in toggle in Provider Settings, I'm happy to follow up with that instead.

## Testing

- Verified against a live Ollama OpenAI-compatible endpoint: with the full prompt the model narrated JSX and never called tools; with the compact prompt it calls `render` and the design appears on the canvas.
- Headless-validated generated JSX through `renderJSX` to confirm the guardrails remove the throwing cases (bad icon names, undefined components).
- `oxlint` clean; `check:i18n` in sync (no i18n changes); app builds/serves.

## Files

- `src/app/ai/chat/system-prompt-compact.md` (new)
- `src/app/ai/chat/transports.ts` — pick the compact prompt for compatible providers

## Summary by CodeRabbit

* **Improvements**
  * Optimized AI assistant performance in the vector design editor with provider-specific configurations to enhance tool execution consistency and reduce unnecessary text output during design sessions.

### DIFF
--- a/src/app/ai/chat/system-prompt-compact.md
+++ b/src/app/ai/chat/system-prompt-compact.md
@@ -1,0 +1,74 @@
+You are a design assistant inside a vector design editor. You act ONLY by calling tools.
+
+🚫 NEVER write JSX, code blocks, markdown, or a plan as a text reply. Text output does NOT draw anything on the canvas — only tool calls do.
+✅ To create or change ANY design, immediately call the `render` tool with a JSX string. Do not explain first. Do not output the JSX as text.
+✅ Your only allowed text reply is a final **1–2 line** summary AFTER you have finished calling tools (frame size, accent color hex).
+
+If the user asks for a screen/UI, your FIRST action must be a `render` tool call.
+
+# Rendering
+
+The `render` tool takes JSX and produces design nodes. JavaScript expressions (map, ternaries, Array.from) work inside JSX. **Each render call must have exactly ONE root element.** To add multiple siblings to the same parent, use separate render calls.
+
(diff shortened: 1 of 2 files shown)


seanshin:
Closing — moving this work to a standalone personal repository. Thanks!

PR #363, opened 2026-06-24
Gray canvas background, Frame tool presets, and architecture docs

JuperoDev:
## Summary
- Change the default canvas/page background from near-white (`#f5f5f5`) to a visibly gray `#e5e5e5` (`packages/core/src/constants.ts`).
- Add a "Frame presets" section to the properties panel, shown while the Frame tool is active — categorized common sizes (desktop, tablet, phone, watch); clicking a preset creates a sized `FRAME` node. Includes i18n strings for all locales and an e2e test.
- Add `ARCHITECTURE.md`: file-by-file documentation of the frontend (`src/` desktop app and `packages/vue/src` SDK), plus module-level summaries of the `core`, `cli`, `mcp`, and `docs` packages.

## Test plan
- [ ] `bun run test:e2e tests/e2e/toolbar/basic.spec.ts` passes, including the new "Frame tool shows frame presets..." test
- [ ] Manually verify canvas background renders gray instead of white on a new page
- [ ] Manually verify Frame tool panel shows presets and clicking one creates a correctly sized frame
- [ ] Spot-check translated strings in `de`, `es`, `fr`, `it`, `pl`, `ru`, `zh-CN` locale files

## Summary by CodeRabbit

* **New Features**
  * Added frame preset options for quickly creating common desktop, tablet, phone, and watch canvases.
  * Expanded language support so the new frame preset labels appear in more locales.

* **Documentation**
  * Refreshed architecture documentation for the app’s structure and quality checks.

* **Bug Fixes**
  * Updated the default canvas background to a slightly darker shade.


dannote:
Thanks for the PR. Frame presets are worth adding, but I don’t think we should merge this branch as it stands.

The background change moves us away from Figma’s current `#F5F5F5` default, and the architecture document duplicates documentation we already maintain elsewhere.

I also checked the preset UI in the current Figma desktop app. The catalog and behavior have changed enough that this implementation would need to be rebuilt: presets remain visible when another layer is selected, categories collapse, new frames use the preset name, and existing frames can be resized from a preset dropdown.

Since the branch now conflicts with master, I think it’s better to close this and revisit frame presets separately on top of the current property panel. I opened #412 to track that work. Thanks for identifying the missing workflow.

PR #381, opened 2026-07-04
Add Requesty as an AI provider

Thibaultjaigu:
Adds Requesty as an AI provider, mirroring the existing OpenRouter provider.

Requesty (https://requesty.ai) is an OpenAI-compatible LLM router — base URL `https://router.requesty.ai/v1`, `GET /v1/models` (OpenAI shape) for live model discovery, `provider/model` naming, optional `HTTP-Referer`/`X-Title` headers.

Changes:
- `packages/core/src/constants.ts`: `'requesty'` in the `AIProviderID` union + a `requesty` entry in `AI_PROVIDERS` (mirrors OpenRouter; verified model ids; appended after OpenRouter so the default provider/model index is unchanged).
- `src/app/ai/chat/model.ts`: `requesty` in the model resolver + a `createLanguageModel` case using the OpenAI-compatible client with base URL `https://router.requesty.ai/v1`.
- `src/app/ai/chat/provider-models.ts`: live `/v1/models` listing + curated fallback, mirroring OpenRouter's.
- `src/app/ai/chat/transports.ts`: Requesty `anthropic/*` models get the same Anthropic cache-control as OpenRouter's.
- `src/components/chat/ProviderSetup.vue` + `ProviderSettings/CustomEndpointSection.vue`: Requesty behaves like OpenRouter (curated list, optional custom model). The provider dropdown is data-driven from `AI_PROVIDERS`, so no dropdown change was needed.

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.

## Summary by CodeRabbit

* **New Features**
  * Added support for a new AI provider option in the app.
  * Model suggestions now work for this provider, making it easier to pick a compatible model.

* **Bug Fixes**
  * Improved model loading and fallback behavior so available models are shown more reliably.
  * Enhanced compatibility for certain Anthropic-based models when used through the new provider.

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -174,6 +174,7 @@ export const ACP_AGENTS: ACPAgentDef[] = [
 
 export type AIProviderID =
   | 'openrouter'
+  | 'requesty'
   | 'anthropic'
   | 'openai'
   | 'google'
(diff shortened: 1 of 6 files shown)


dannote:
Thanks for the PR and for being transparent about the Requesty affiliation.

We’re going to pass on adding Requesty as a first-class built-in provider for now. OpenPencil already supports OpenAI-compatible/custom endpoints, and we don’t want the
provider list to become a catalog of router services unless there is clear user demand or a materially distinct integration need.

This looks like it can be configured through the existing OpenAI-compatible provider path, so we’d rather keep the built-in list focused.

Closing as not planned.

PR #388, opened 2026-07-10
fix: multi-color vector fills on .fig import (e.g. FedEx logo)

### LINKED ISSUE #386: Multi-color vector fills ignored on .fig import (e.g. FedEx logo)
## Summary

Some `.fig` files that use **multi-color vector fills** (Figma `fillGeometry[].styleID` + `vectorData.styleOverrideTable`) are imported and rendered incorrectly: every path is painted with the node's single `fills` paint.

Classic repro: **FedEx logo** — “Ex” should be orange, but OpenPencil shows the whole wordmark purple.

**Fix PR:** https://github.com/open-pencil/open-pencil/pull/388

### Incorrect (all purple)

FedEx incorrect — entire wordmark purple

### Correct (purple Fed + orange Ex)

FedEx correct — purple Fed, orange Ex

## Expected

- Multi-region vectors keep **per-path fills** from Figma style overrides.
- FedEx logo: purple “Fed…” and **orange “Ex”**.

## Actual

- All logo paths use the node-level purple fill.
- Path-level orange override (`styleID: 1` → orange solid in `styleOverrideTable`) is ignored when painting.

## Root cause

1. **Import:** `resolveGeometryPaths` drops `styleID` and never resolves `styleOverrideTable[].fillPaints` onto paths.
2. **Render:** VECTOR fill drawing applies `node.fills` to **every** `fillGeometry` path.

### Technical detail

- Node has `fillPaints` (e.g. purple) as default style `0`.
- `vectorData.styleOverrideTable` holds extra styles (e.g. style `1` = orange solid).
- `fillGeometry[]` entries reference `styleID` per path (e.g. `[1, 1, 0]`).
- Each path must be painted with the fills for its `styleID` (0 → node fills).

## Suggested fix

- Preserve `styleID` / optional `fills` on `GeometryPath`.
- At import, map `styleOverrideTable` paints onto paths.
- At paint time, draw VECTOR multi-style fills per path.

### Relevant code

- `packages/scene-graph/src/types.ts` — `GeometryPath`
- `packages/core/src/kiwi/fig/node-change/vector-geometry.ts` / `convert.ts`
- `packages/core/src/canvas/fills.ts` / `scene.ts`
- `tests/engine/kiwi/vector-style-fills.test.ts`

## Repro fixture

- [FedEx.fig.zip](https://github.com/user-attachments/files/29870955/FedEx.fig.zip) — unzip to get `FedEx.fig`

Open via **Open local .fig…** or drag onto the editor.

## Repro

1. Open a `.fig` with multi-color vector fills (e.g. FedEx logo export).
2. Observe path colors collapse to a single fill (all purple).

rcoenen:
### Summary

Some `.fig` files with multi-color vectors render every path with a single fill. Classic case: FedEx logo shows all purple instead of purple + orange “Ex”.

Fixes #386

### What changed

- Import preserves `fillGeometry` `styleID` and resolves paints from `vectorData.styleOverrideTable` onto each path
- VECTOR painting draws per-path fills when path-level fills exist (`drawVectorMultiStyleFills`)
- Regression test for multi-style wordmark fills

### Validation

- [x] `bun test ./tests/engine/kiwi/vector-style-fills.test.ts`
- [x] Manual: open [FedEx.fig](https://github.com/user-attachments/files/29870955/FedEx.fig.zip) and confirm purple + orange

#### Before (all purple)

FedEx before — entire wordmark purple

#### After (purple Fed + orange Ex)

FedEx after — purple Fed, orange Ex

## Summary by CodeRabbit

- **New Features**
  - Added multi-style vector fill rendering that supports per-path fill overrides, including correct handling of visibility, per-fill alpha, and blend behavior.
- **Bug Fixes**
  - Improved vector fill styling so path-level style overrides take priority over node-level fills when present.
  - Ensured per-path style identifiers and resolved fills are preserved during geometry copying and FIG export/import.
- **Tests**
  - Added vector fill style override tests, including resolution and end-to-end export/import roundtrip checks.


packages/core/src/kiwi/fig/node-change/vector-geometry.ts
```diff
@@ -58,10 +82,17 @@ export function resolveGeometryPaths(
...
+    const styleID = typeof p.styleID === 'number' ? p.styleID : undefined
+    const path: GeometryPath = {
       windingRule: (p.windingRule === 'EVENODD' ? 'EVENODD' : 'NONZERO') as WindingRule,
       commandsBlob: blob
-    })
+    }
+    if (styleID != null && styleID !== 0) {
+      path.styleID = styleID
+      const fills = styleFillsById?.get(styleID)
+      if (fills && fills.length > 0) path.fills = fills
```
dannote, line 93: Could we carry this through `.fig` serialization as well? The import now stores `styleID` and `fills`, and copying preserves them, but `serializeGeometry()` still drops the path `styleID` and rebuilds `styleOverrideTable` only from handle mirroring. Once an imported vector is edited and its raw metadata is cleared, exporting and reopening it loses the multiple colors. It would be good to emit the path style IDs and their `fillPaints` into a unified override table, with an export → import test.

rcoenen: Done in a05eb00e — `serializeGeometry()` now emits each path's `styleID` and merges its `fillPaints` into the unified `styleOverrideTable` (handle-mirroring entries keep IDs 1..n, fill entries continue after), and `vector-style-fills.test.ts` gains an export → import roundtrip test asserting the per-path colors survive re-serialization.

packages/core/src/canvas/fills.ts
```diff
@@ -3,9 +3,44 @@ import type { Canvas, Paint } from 'canvaskit-wasm'
...
+      if (!fill?.visible) continue
+      if (!applyFill(r, fill, node, graph, fi)) continue
+      r.fillPaint.setAlphaf(fill.opacity)
+      r.fillPaint.setBlendMode(figmaBlendModeToSkia(r.ck, fill.blendMode))
+      canvas.drawPath(path, r.fillPaint)
+      r.fillPaint.setShader(null)
+      r.fillPaint.setBlendMode(r.ck.BlendMode.SrcOver)
+    }
+    path.delete()
+  }
```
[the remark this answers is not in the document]
rcoenen: Done in fb96fb69 — `drawVectorMultiStyleFills` now uses `r.getFillGeometry(node)` (cached `Path[]`, no per-paint allocation/delete).

packages/core/src/canvas/fills.ts
```diff
@@ -3,9 +3,44 @@ import type { Canvas, Paint } from 'canvaskit-wasm'
...
+    for (let fi = 0; fi < fills.length; fi++) {
+      const fill = fills[fi]
+      if (!fill?.visible) continue
+      if (!applyFill(r, fill, node, graph, fi)) continue
+      r.fillPaint.setAlphaf(fill.opacity)
+      r.fillPaint.setBlendMode(figmaBlendModeToSkia(r.ck, fill.blendMode))
+      canvas.drawPath(path, r.fillPaint)
+      r.fillPaint.setShader(null)
+      r.fillPaint.setBlendMode(r.ck.BlendMode.SrcOver)
+    }
```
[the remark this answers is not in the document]
rcoenen: Done in fb96fb69 — extracted `paintFills(r, fills, node, graph, draw)` in `fills.ts`; both `drawVectorMultiStyleFills` and `drawVisibleFills` in `scene.ts` now delegate to it.

packages/core/src/kiwi/fig/node-change/convert.ts
```diff
@@ -504,9 +508,15 @@ function convertVectorAndStrokeProps(nc: NodeChange, blobs: Uint8Array[]) {
   const vectorNetwork = resolveVectorNetwork(nc, blobs)
   const strokeCap = getVectorStrokeCap(nc, vectorNetwork)
   const strokeJoin = getVectorStrokeJoin(nc, vectorNetwork)
+  // Multi-color vectors (e.g. FedEx purple + orange) store per-path paints on
+  // vectorData.styleOverrideTable, referenced by fillGeometry[].styleID.
+  const vectorData = nc.vectorData as
+    | { styleOverrideTable?: Parameters<typeof styleOverrideFillsById>[0] }
+    | undefined
```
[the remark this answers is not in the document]
rcoenen: Done in fb96fb69 — `VectorStyleOverride` is imported directly and used for the `vectorData` cast.

rcoenen:
@dannote All review feedback is addressed and CodeRabbit has approved. Heads-up on the remaining red CI: the `quality` and `Unit tests (*)` jobs all fail in setup at `git lfs pull` with `LFS: Authorization error` against the R2 proxy — fork PRs don't get the LFS credentials, so they die before any tests run ([example log](https://github.com/open-pencil/open-pencil/actions/runs/29126309520/job/86472545999)). Locally the affected suites pass. The earlier preview failure was a real `format:check` miss in `renderer.ts`, fixed in 18833857.

rcoenen:
@coderabbitai review

packages/scene-graph/src/types.ts
```diff
@@ -72,6 +72,13 @@ export interface VectorNetwork {
 export interface GeometryPath {
   windingRule: WindingRule
   commandsBlob: Uint8Array
+  /**
+   * Figma fillGeometry styleID. 0 / omitted → use the node’s fills.
+   * Non-zero → paints from vectorData.styleOverrideTable for multi-color vectors.
+   */
+  styleID?: number
```
dannote, line 79: Could we keep the Figma `styleID` out of the normalized `GeometryPath` model? The generic scene data only needs the resolved per-path `fills`; Figma IDs are transport metadata that can be consumed on import and regenerated on export. Keeping the ID as scene state is what allows it to become stale after editing.

packages/core/src/kiwi/fig/node-change/serialize.ts
```diff
@@ -409,24 +410,38 @@ function serializeGeometry(node: SceneNode, nc: KiwiNodeChange, blobs: Uint8Arra
     nc.maskType = node.maskType
     if (node.maskIsOutline) nc.maskIsOutline = true
   }
+  // Re-emit per-path fill overrides (multi-color vectors, e.g. FedEx purple +
+  // orange) so path colors survive export → import once raw metadata is gone.
+  const pathStyleIds = new Map<number, number>() // imported styleID → exported styleID
```
dannote, line 415: Using the imported `styleID` as the export key can lose current scene data. I reproduced this with two paths sharing an original ID but having different fills: after export → import, both paths received the first fill. Path fills are also dropped entirely when `vectorNetwork` is null because this map is only populated inside that block. Could we derive fresh style IDs from the current path fills, independently of the imported ID and vector network?

packages/core/src/kiwi/fig/instance-overrides/constraints.ts
```diff
@@ -63,16 +63,29 @@ function scaleGeometryBlobs(geom: GeometryPath[], sx: number, sy: number): Geome
...
+        if (offset + 8 > scaled.length) break
         dv.setFloat32(offset, dv.getFloat32(offset, true) * sx, true)
         dv.setFloat32(offset + 4, dv.getFloat32(offset + 4, true) * sy, true)
         offset += 8
       }
     }
-    return { windingRule: g.windingRule, commandsBlob: scaled }
+    // Keep styleID + path-level fills (multi-color vectors). Dropping them makes
+    // constraint-scaled instances paint as a single node fill.
+    return {
```
dannote, line 83: There is another `scaleGeometryBlobs()` in `derived-symbol-data/geometry.ts` that still returns only `windingRule` and `commandsBlob`; I confirmed that path drops the new fills during scaling. Could we extract one shared geometry-blob scaling helper and use it in both places, so opcode handling and copied metadata stay consistent?

packages/core/src/canvas/fills.ts
```diff
@@ -16,6 +67,8 @@ export function drawNodeFill(
 ): void {
   switch (node.type) {
     case 'VECTOR': {
+      // When path-level fills exist, multi-style drawing is handled separately.
+      if (node.fillGeometry.some((g) => g.fills && g.fills.length > 0)) break
```
dannote, line 71: This works through `renderShapeUncached()`, but `drawNodeFill()` is also called directly by `recordPatternSource()`. In that path the multi-style dispatcher is never reached, so a multi-color vector used as a pattern source draws nothing. Could we centralize the vector fill dispatch so all callers get the same behavior?

tests/engine/kiwi/vector-style-fills.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    expect(paths[2]?.styleID).toBeUndefined()
+    expect(paths[2]?.fills).toBeUndefined()
+  })
+
+  describe('export → import roundtrip', () => {
+    beforeAll(async () => {
+      await initCodec()
+    })
+
+    test('per-path fill overrides survive serialization', async () => {
```
dannote, line 99: Could we add a canvas snapshot for this? These tests cover normalized state and a synthetic export/import roundtrip, but they never run the renderer. Since this changes pixel output, the repository’s renderer changes normally include visual coverage for the actual per-path colors.

dannote (changes requested):
Thanks for working through this. I left a few comments around the scene-model boundary, geometry transformations, rendering dispatch, and visual coverage. The changed files also currently fail type-aware lint.

dannote:
Thanks, Rob! This has been adapted and landed on `master` as `a7d7b719`, with your co-author credit preserved. I’m closing the superseded PR.

PR #390, opened 2026-07-10
fix(vector-edit): make double-click vector editing work on opened files

rcoenen:
## Problem

Double-clicking a vector in any **opened** document never entered vector edit mode — and once that first bug is fixed, three more surface behind it: nested vectors edited at the wrong coordinates, colored fills didn't follow edits, and undo didn't work.

Repro: import any .fig with a vector inside a frame (e.g. a multi-color logo), double-click through to the vector, double-click again → nothing happens (Figma shows anchor points).

## Root causes & fixes

1. **Stale graph reference** — `createVectorEditActions(editor, graph, state)` captured the graph at store creation, but opening a file swaps the instance via `editor.replaceGraph()`. `enterNodeEditMode` looked nodes up in the dead startup graph and silently returned. The vector-edit and pen modules now read `editor.graph` (live getter) at call time.

2. **Parent-relative vs page-absolute coordinates** — edit-session vertices were built with `v + node.x/y`, which is parent-relative. For a vector nested in a frame, the live overlay drew the shape and anchors off-screen (the node itself is hidden during editing → blank canvas). Sessions now map through the node's full **world matrix** (vertices as points, tangents as direction vectors), so nested, rotated, and flipped vectors edit in place. Commit converts back with rotation-preserving center math; the live overlay renders through `canvas.concat(worldMatrix)`. Pen resume gets the same treatment. New shared helper: `transformVectorNetwork(m, network)`.

3. **Fills didn't follow edits** — vector fills render from static imported `fillGeometry` command blobs, so network edits changed the outline data but not the visible fills (multi-color *and* single-color). Imported .fig vectors pair `fillGeometry[i]` with `regions[i]` positionally (verified empirically: start vertices, loop counts, and winding rules match). `regenerateFillGeometry()` re-emits each region's path commands from the edited network — on commit and per-frame in the live overlay — preserving per-entry styleID/fills. Regenerated blobs are in the .fig format, so edits also round-trip back to Figma. If region count no longer matches (e.g. a region deleted mid-edit) it keeps the old blobs rather than guess.

4. **Undo** — commit previously wrote via `graph.updateNode()`, bypassing the undo stack. It now goes through `updateNodeWithUndo` (one "Edit vector" entry). Additionally, a session-local history handles **Cmd+Z / Cmd+Shift+Z inside edit mode**: geometry snapshots are pushed at drag start and before each discrete op (delete/add/break/connect/align/zero-handles); `edit.undo`/`edit.redo` route to it while `nodeEditState` is active and fall back to the document stack otherwise.

Includes the two multi-color fillGeometry commits this builds on (`honor multi-color vector fillGeometry style overrides` + its unit test).

## Testing

- 9 new unit tests: world-matrix round-trips with rotated frame + rotated vector (enter/commit/no-op invariants, tangent rotation), fillGeometry regeneration (positional rebuild, styleID/fills preservation, cubics, mismatch bail, regionless chains), session undo/redo (restore, redo, no-op-snapshot skipping, empty-stack no-op)
- Verified end-to-end in the browser with a real multi-color .fig: double-click enters edit mode with anchors on the artwork (also under a 30°-rotated parent), fills track vertex drags live and after commit, Cmd+Z works both inside the session and at document level after commit
- `oxlint`, `oxfmt`, and `tsgo --noEmit` clean

## Summary by CodeRabbit

* **New Features**
  * Added per-path multi-style vector fills with correct paint/opacity/blend handling.
  * Added per-path style override support (styleID→fills) and updated export/import to preserve path-level fill overrides.
  * Added vector fill-geometry regeneration, plus shared helpers for scaling/copying geometry paths.
  * Enhanced vector editing rendering and added geometry-based session undo/redo and edit-history hooks.
* **Bug Fixes**
  * Fixed VECTOR fill rendering fallback to ensure per-path overrides are selected when available.
  * Corrected live node edit overlay geometry by using full world-matrix transforms.
* **Tests**
  * Added unit, engine, and e2e coverage for multi-style fills, geometry regeneration, transforms, undo/redo, and visual snapshots.

## Screenshot

Vector edit mode on an imported multi-color .fig — every anchor point of the network rendered in place over the artwork:

Multi-color vector in edit mode with anchor points aligned on the artwork


src/app/editor/vector-edit/lifecycle.ts
```diff
@@ -1,70 +1,86 @@
...
+      node.id,
+      {
+        x: center.x - bounds.width / 2,
+        y: center.y - bounds.height / 2,
+        width: bounds.width,
+        height: bounds.height,
+        vectorNetwork: relativeNetwork,
+        // Fills render from fillGeometry blobs when present — rebuild them from
+        // the edited network so fills follow the edit.
+        fillGeometry: regenerateFillGeometry(relativeNetwork, node.fillGeometry)
```
dannote: I think we are still losing the per-path colors when an edited vector is exported. `regenerateFillGeometry` keeps `styleID` and `fills` in memory, but `serializeGeometry()` writes only `windingRule` and `commandsBlob`, and rebuilds `styleOverrideTable` only from handle mirroring. Since this update invalidates the imported raw geometry metadata, the new paths are exported without their path-level paints. Could we serialize these style IDs and fill paints into the unified override table and add an export → import test for a multi-color vector?

src/app/editor/vector-edit/history.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { VectorNetwork } from '@open-pencil/scene-graph'
+
+import type { NodeEditState, VectorEditState } from './types'
+
+function snapshot(es: NodeEditState): VectorNetwork {
+  return cloneVectorNetwork({ vertices: es.vertices, segments: es.segments, regions: es.regions })
+}
+
+function sameGeometry(a: VectorNetwork, b: VectorNetwork) {
+  return JSON.stringify(a) === JSON.stringify(b)
```
dannote: Could we use structural equality here instead of comparing serialized strings? `JSON.stringify` depends on property order, has surprising behavior for a few values, and allocates two complete strings on every comparison. `isEqual` from `es-toolkit/predicate` would be a better fit, either declared for the app package or exposed through a semantic scene-graph helper.

src/app/editor/vector-edit/history.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/** Snapshot the current geometry onto the session undo stack (call BEFORE mutating). */
+export function pushNodeEditHistory(es: NodeEditState) {
+  es.future = []
```
dannote: This clears redo before we know whether anything changed. The history snapshot is pushed on pointer-down, so after undoing, simply clicking a vertex without dragging removes the redo history. Could we wait until the first actual geometry change before clearing `future`? A small undo → click → redo test would cover it.

src/app/editor/vector-edit/lifecycle.ts
```diff
@@ -1,70 +1,86 @@
...
+    // The node rotates/flips about its center, so with rotation preserved the
+    // new x/y follow from where the new geometry's center lands in the parent
+    // frame (via the OLD local matrix — the geometry itself has not moved).
+    const localMatrix = getNodeLocalMatrix(node)
+    const center = Matrix.mapPoint(localMatrix, {
+      x: bounds.x + bounds.width / 2,
+      y: bounds.y + bounds.height / 2
     })
+
+    editor.updateNodeWithUndo(
```
dannote, line 59: This also creates an `Edit vector` undo entry when someone enters edit mode and exits without changing anything. The current no-op test checks the geometry but not the undo stack. Could we skip the update when the resulting bounds/network/fill geometry are unchanged and assert that the no-op session does not enable document undo?

tests/engine/kiwi/vector-style-fills.test.ts
```diff
@@ -0,0 +1,78 @@
+import { describe, expect, test } from 'bun:test'
+
+import {
+  resolveGeometryPaths,
+  styleOverrideFillsById
+} from '../../../packages/core/src/kiwi/fig/node-change/vector-geometry'
```
dannote: This import currently fails the repository lint rules for both cross-package source imports and deep parent-relative imports. The nearby `.fig` tests use `#core/...`, so `#core/kiwi/fig/node-change/vector-geometry` should work here too. The non-null assertions below also need fixing, as noted in the existing review.

packages/core/src/canvas/node-edit-overlay.ts
```diff
@@ -124,55 +126,40 @@ function drawLiveShape(
...
-  // renderShapeUncached expects a canvas translated to the node's local origin.
+  // The overlay canvas is in screen space; renderShapeUncached expects the
+  // node's local frame, so apply viewport then the node's world matrix.
   canvas.save()
-  canvas.translate(bounds.x * r.zoom + r.panX, bounds.y * r.zoom + r.panY)
+  canvas.translate(r.panX, r.panY)
   canvas.scale(r.zoom, r.zoom)
+  canvas.concat(world)
 
   r.renderShapeUncached(canvas, node, graph)
```
dannote: Could we put the temporary node changes and render call behind `try/finally`? If rendering throws, `canvas.restore()` is skipped and the node keeps the live network/fill geometry. The cleanup should always restore the canvas, original node properties, and caches.

dannote (changes requested):
Thanks for digging into this — the stale graph and world-transform fixes make sense, and the tests are thoughtful. There are a few issues to address before merging, mainly `.fig` export of multi-color paths, undo edge cases, and lint. Since this changes vector rendering, it also needs canvas snapshot coverage.

src/app/editor/vector-edit/lifecycle.ts
rcoenen: This is now inherited from #388 — the branch is rebased onto it, so `serializeGeometry()` emits path `styleID`s and merges their `fillPaints` into the unified `styleOverrideTable`, with an export → import roundtrip test in `vector-style-fills.test.ts`.

src/app/editor/vector-edit/history.ts
rcoenen: Done in 1bf8c802 — added a semantic `vectorNetworksEqual` helper to scene-graph (via `isEqual` from `es-toolkit/predicate`, already a scene-graph dependency) and the history module uses it instead of `JSON.stringify`.

src/app/editor/vector-edit/history.ts
rcoenen: Done in 1bf8c802 — pointer-down pushes without touching `future`; stale redo entries are invalidated lazily when the geometry diverges from the last undo/redo restore point. Added both tests: undo → click → redo keeps the redo, and a real change after undo clears it.

src/app/editor/vector-edit/lifecycle.ts
rcoenen: Done in 1bf8c802 — the session stores the geometry exactly as entered (`origAbsNetwork`) and commit early-returns when the live network is structurally identical, so a no-op session never calls `updateNodeWithUndo`. The no-op test now asserts the undo stack stays empty.

tests/engine/kiwi/vector-style-fills.test.ts
rcoenen: Fixed on the #388 branch (b8d0cb4f, inherited here via rebase) — the test imports through `#core/kiwi/fig/node-change/vector-geometry` and the non-null assertions were replaced with the `expectDefined` helper.

packages/core/src/canvas/node-edit-overlay.ts
rcoenen: Done in 1bf8c802 — the transform + render now sit in `try/finally`; canvas restore, node property restore, and cache invalidation always run.

packages/core/src/canvas/fills.ts
```diff
@@ -3,9 +3,44 @@ import type { Canvas, Paint } from 'canvaskit-wasm'
...
+ * Paths without path.fills use node.fills. Returns true when handled.
+ */
+export function drawVectorMultiStyleFills(
+  r: SkiaRenderer,
+  canvas: Canvas,
+  node: SceneNode,
+  graph: SceneGraph
+): boolean {
+  if (node.type !== 'VECTOR' || node.fillGeometry.length === 0) return false
+  const hasPathFills = node.fillGeometry.some((g) => g.fills && g.fills.length > 0)
```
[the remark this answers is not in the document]
rcoenen: Done in 4cc0aac2 — extracted `hasPathLevelFills(node)` used by both call sites.

tests/engine/kiwi/vector-style-fills.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    ])
+
+    expect(map.has(1)).toBe(true)
+    expect(map.has(2)).toBe(false)
+    const orange = map.get(1)?.[0]?.color
+    expect(orange).toBeDefined()
+    expect(orange!.r).toBeCloseTo(1, 2)
+    expect(orange!.g).toBeCloseTo(0.32, 2)
+    expect(orange!.b).toBeCloseTo(0, 2)
```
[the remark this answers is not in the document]
rcoenen: Fixed on the #388 branch (inherited here via rebase) — the assertions now go through the `expectDefined` helper, no `!` remains in the file.

rcoenen:
@dannote All review points are addressed — thanks for the thorough pass. Summary:

- **.fig export of edited multi-color vectors**: the branch is rebased onto #388, which now serializes path `styleID`s + `fillPaints` into the unified `styleOverrideTable` with an export → import roundtrip test. Once #388 merges this PR shrinks to just the vector-edit commits.
- **Structural equality** (1bf8c802): new `vectorNetworksEqual` scene-graph helper (es-toolkit `isEqual`) replaces the `JSON.stringify` comparison.
- **Redo clearing** (1bf8c802): pointer-down no longer clears `future`; stale redo entries are invalidated only once geometry actually diverges from the last restore point. Tests: undo → click → redo keeps redo; a real change after undo clears it.
- **No-op sessions** (1bf8c802): commit early-returns when the session geometry is unchanged, so enter + exit never creates an `Edit vector` undo entry — asserted in the no-op test.
- **Overlay cleanup** (1bf8c802): `drawLiveShape` restores canvas, node props, and caches in `try/finally`.
- **Canvas snapshot coverage** (f68d2c65): new `vector-edit-visuals.spec.ts` with goldens for multi-color per-path fills (flat + nested in a rotated frame) and the edit-mode live overlay on a nested rotated vector.
- Lint: the test now imports via `#core/...` and uses `expectDefined` instead of `!` (from #388), plus `hasPathLevelFills` extraction (4cc0aac2).

tests/engine/vue/input/resize-fill-geometry.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(Buffer.from(geo.commandsBlob).equals(Buffer.from(origBlob))).toBe(false)
+    // scaled blob: second command's x coordinate is the new width
+    expect(
+      new DataView(geo.commandsBlob.buffer, geo.commandsBlob.byteOffset).getFloat32(10, true)
+    ).toBeCloseTo(50, 3)
+    // styleID and per-path fill survive the rebuild
+    expect(geo.styleID).toBe(1)
+    expect(geo.fills?.[0]?.color.r).toBeCloseTo(1, 2)
+  })
+})
```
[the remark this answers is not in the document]
rcoenen: Done in 58bf72ab — the commit-path test now asserts the persisted fillGeometry carries the scaled blob with styleID/fills intact, and that the undo payload restores the original width and blob bytes (on top of the structured-clone safety check added in 53a54645).

tests/engine/vue/input/resize-fill-geometry.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(
+      new DataView(finalGeo.commandsBlob.buffer, finalGeo.commandsBlob.byteOffset).getFloat32(
+        10,
+        true
+      )
+    ).toBeCloseTo(50, 3)
+    expect(finalGeo.styleID).toBe(1)
+    expect(finalGeo.fills?.[0]?.color.r).toBeCloseTo(1, 2)
+
+    // …and the undo payload carries the ORIGINAL geometry for rollback
```
dannote, line 150: This verifies the original payload using a mocked `commitResize()`, but it does not exercise the real undo manager. Core `createResizeSnapshot()` still omits `fillGeometry`, so redo is currently broken: after resizing, undo restores the original network and blob, but redo restores only the scaled `vectorNetwork` and leaves the original `fillGeometry`.

I reproduced this with `createEditor()`: the final blob represented size 20, undo restored 10, and redo left the blob at 10 while the network returned to 20. `commitGroupResize()` has the same problem because its final child snapshots also omit `fillGeometry`.

Please include `fillGeometry` in the core resize snapshots and add real `createEditor()` tests covering resize → undo → redo for both a root vector and a vector inside a resized group.

packages/vue/src/shared/input/resize.ts
```diff
@@ -21,7 +24,14 @@ function resizeChanges(d: DragResize, cx: number, cy: number, constrain: boolean
     newRect.width,
     newRect.height
   )
-  if (resizedVectorNetwork) changes.vectorNetwork = resizedVectorNetwork
+  if (resizedVectorNetwork) {
+    changes.vectorNetwork = resizedVectorNetwork
+    // Vectors render from fillGeometry blobs when present — rebuild them from
+    // the scaled network or the artwork stays at its old size.
+    if (d.origFillGeometry.length > 0) {
```
dannote, line 31: This fixes `fillGeometry`, but imported vector strokes can also render from `strokeGeometry`. Those blobs are not scaled or cleared here, so resizing an outlined imported vector changes its network and bounds while its rendered stroke outline remains at the old size.

Please track the original `strokeGeometry`, scale it with the existing geometry-path scaling helper, persist it through commit and undo/redo, and add a regression test alongside the fill-geometry coverage.

dannote (changes requested):
Thanks for working through the earlier feedback. I verified those fixes, and the targeted vector-edit and visual tests now pass. I found one remaining resize-history bug and a related stroke-geometry gap; this branch also still depends on the unresolved #388 changes.

dannote:
Thanks, Rob! This has been adapted and landed on `master` as `cb2c4aa3`, with your co-author credit preserved. I’m closing the superseded PR.

PR #389, opened 2026-07-10
feat(app): optional BYOK S3 cloud storage with local-first sync

rcoenen:
### Summary

Optional user-owned S3-compatible cloud storage (B2, AWS, R2, MinIO, …) with a Files home when configured. Unconfigured installs keep local-only behavior.

Saves and imports write to a durable local cache first and sync to the bucket in the background so large uploads do not block the editor.

### What changed

- Cloud storage settings (credentials, Test connection, CORS helpers)
- Files home: list / create / open / rename / delete / open local / drag-drop import
- Namespace `open_pencil_storage/` for all objects
- Local-first store + outbox sync (pending/offline/error status)
- Import validation, unique names, thumbnails
- **File → New** (and new-tab) creates a cloud canvas when cloud is configured
- OpenSpec + docs + unit tests under `tests/engine/cloud/**`

### Validation

- [x] `bun test ./tests/engine/cloud` (35 pass)
- [ ] `bun run check` (recommend CI)
- [x] CHANGELOG.md updated
- [ ] Manual B2 smoke: list, import, open, save, refresh, offline pending

### Notes

- Browser needs bucket CORS; desktop Test connection can apply CORS when the key allows.
- Multi-color vector import (#386 / #388) is a separate PR — not part of this change.

## Summary by CodeRabbit

* **New Features**
  * Added optional S3-compatible cloud storage with a Files home for creating, opening, renaming, deleting, and importing canvases.
  * Added local-first saving with background synchronization, offline support, retry handling, thumbnails, and visible sync statuses.
  * Added cloud storage settings, connection testing, CORS guidance, usage details, and secret visibility controls.
  * Added support for per-path vector fill styles.
* **Bug Fixes**
  * Improved startup loading feedback and ensured local documents remain available while cloud synchronization is pending.
* **Documentation**
  * Added cloud storage setup guidance and updated the changelog and roadmap.

PR #391, opened 2026-07-10
feat: introduce OpenPencil's Cloud Workspace (BYOK S3 sync)

rcoenen:
### OpenPencil, meet your workspace

Until now, OpenPencil treated design files the way desktop apps always have: one `.fig` on disk, one window, and it's on you to remember where you put it. The thing that makes Figma feel like a *place* rather than a program — the home screen full of live thumbnails that follows you to any machine — was the missing piece for daily work.

This PR introduces the **Cloud Workspace**: that same workspace experience, built the OpenPencil way — **bring your own key**. There is no OpenPencil account and no middleman server. Point the app at any S3-compatible bucket (AWS S3, Backblaze B2, Cloudflare R2, MinIO, …) and your designs follow you across devices while every byte stays under your own keys. We test against Backblaze B2, whose 10 GB free tier makes trying this cost exactly nothing. And if you never configure it, nothing changes: local-only mode is untouched.

# Meet the new workspace 
Cloud Workspace

### What's inside

**The workspace** — a project home with live thumbnails, right-click Open / Rename / Duplicate / Download .fig / Delete, proper in-app confirmation and prompt dialogs (native `confirm()`/`prompt()` don't work in Tauri), and a footer connection indicator. A fresh bucket greets you with a Welcome project.

**BYOK storage adapters** — the S3-compatible adapter (`src/app/cloud/s3/`, aws4fetch-signed, path-style URLs, namespaced under `open_pencil_storage/` so a bucket can be shared with other data) is the first of a pluggable family: Google Drive, Nextcloud, or WebDAV can slot in later without touching the sync engine.

**Smart sync** — local-first, always. Every save lands in IndexedDB instantly and uploads in the background through a persistent outbox with retry/backoff, revision-guarded uploads, tombstoned deletes, and remote reconciliation that heals orphaned rows without ever stamping unsynced local edits as synced. Uploads report real progress — the card badge fills as the bytes go up (XHR replay of the signed PUT, since `fetch()` can't observe upload progress) — and the local cache self-evicts past 500 MB, least-recently-opened first; unsynced work and open files are never evicted.

## Smart syncing
Smart sync of large files

## New settings menu
**Settings, revamped** — cloud configuration needed a home, so Settings became one centralized, tabbed dialog hosted consistently top-right: Cloud storage and the existing AI provider configuration now live side by side, ready to grow a tab per future integration.

<p>
Settings — Cloud storage tab
Settings — AI providers tab
</p>

**CORS made painless** — **Test connection** applies web-friendly bucket CORS automatically where permissions allow, or gives a one-click **Copy CORS JSON** for the provider console. Connection failures show a red alert on the Settings button and an inline settings link — never a silently empty workspace.

**Fixes found while testing end-to-end**

- Page background color now persists through save/reload (round-tripped through fig `backgroundColor`, including for OpenPencil-authored pages).
- Stuck "Syncing" indicators (thumbnail writes no longer demote a document's sync status; a failed thumbnail upload no longer marks the whole canvas as errored).
- Delete no longer races reconcile (tombstones survive until the remote listing confirms deletion).
- Assorted UX: header with logo and left-aligned Open .fig / New .fig, copy tweaks, i18n for all new strings across the 9 locales.

### Validation

- [x] `bun run check`
- [x] Tests added or updated: unit tests for the local store, outbox, sync engine behaviors, cache eviction, CORS origin collection, fig thumbnail extraction, and the page-color round-trip
- [x] CHANGELOG.md updated

Everything was additionally verified live against a real Backblaze B2 bucket (multi-device reconcile, delete lifecycle, duplicate/download, a 27 MB upload driving the progress badge, offline → online recovery).

Supersedes #389.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Cloud Workspace (BYOK) for S3-compatible storage: cloud home/cards, right-click actions, previews/thumbnails, and Welcome seeding.
  * Implemented local-first cloud saving with background sync/outbox, retry, pending-sync indicators, and cache eviction.
  * Added Cloud/AI Settings with S3 setup, CORS test/apply + copy guidance, secret show/hide, and cloud routing/edit UI.

* **Bug Fixes**
  * Preserved per-page background color across navigation and FIG round-trips.

* **Documentation**
  * Updated README/changelog and added Cloud storage (S3) guide; clarified BYOK in roadmap.

* **Tests**
  * Added Bun coverage for cloud CORS, sync/outbox, thumbnails, caching/eviction, vector style overrides, and FIG round-trip.


rcoenen:
Note on the red CI: all unit-test jobs are dying in setup with git-LFS authorization errors against the R2 LFS proxy (`git-lfs-s3-proxy.pages.dev/...open-pencil-lfs`). Master's scheduled **Heavy tests** run hit the exact same errors on the same object IDs this morning (run 29075173832), before this PR existed — so the LFS credentials look expired repo-wide rather than anything in this branch (which touches no LFS-tracked files). Happy to re-trigger checks once LFS access is restored.

rcoenen:
Review feedback addressed in 71bf87c2 + 057162db:

**Removed from the PR**
- The `openspec/` change notes are gone from the repo (they were design exploration, and several described an older architecture — all 8 findings on them are moot).

**Fixed**
- All ~65 new Cloud Workspace strings translated in each of the 8 non-English locales (de/es/fr/it/ja/pl/ru/zh-cn); existing partial translations preserved.
- `oxfmt` formatting on `sync/persist.ts` and `validate-import.test.ts`.
- Window drop handler now `preventDefault`s any file drop, so dropping a non-design file can't navigate the browser away.
- `/edit/cloud/:canvasId` gets a `beforeEnter` guard when cloud is unconfigured, and EditorView's catch now always redirects instead of leaving a broken editor.
- Sync engine: a job enqueued mid-pump re-wakes the pump (online only, so no offline spin); `openIdb` rejects on `onblocked`; removed the redundant third `kickSyncEngine` per save.
- `refAutoReset` for the CORS-copied flag; whitespace-only rename is a no-op; prompt input has an `aria-label`; boot-loader fade-out chain has a `.catch` that removes the loader; loader message styles moved into the `<style>` block.
- The random unique-name fallback now loops until the candidate is actually free.
- Bonus found while fixing: `tests/engine/cloud` wasn't in any unit-test shard, so no CI job ran the new cloud tests — added to the `app` group.

**Declined, with reasons**
- *pages.ts pageColor reset*: `restorePageViewport` (called earlier in `switchPage`) already resets `pageColor` to `CANVAS_BG_COLOR` for pages without a saved viewport, so a new page never inherits the previous page's color.
- *outbox TOCTOU / reconcile race / activity scopes / welcome-seed staleness*: real but tiny windows in single-tab usage; the engine's revision/tombstone guards keep outcomes correct, and reconcile is last-write-wins by design. Happy to revisit if multi-tab becomes a target.
- *listCanvases/deleteCanvas partial-failure handling*: deletes retry through the outbox and S3 deletes are idempotent, so orphans self-heal on retry.
- *HomeView nested interactive spans*: valid a11y point; the context menu already provides keyboard-accessible actions, so restructuring the card markup is deferred to a follow-up.
- *credentials clear vs env prefill*: dev-only scenario (`.env.local` prefill); noted for a follow-up.

rcoenen:
All remaining review findings are now addressed in 773eb8d5 — nothing left open:

- **Pattern-source vectors** (fills.ts): `recordPatternSource` now runs the multi-style pass, so vectors with per-path fills no longer record an empty pattern; the multi-style pass also reuses the renderer's cached fill geometry instead of rebuilding Skia paths every frame.
- **Activity label**: replaced with a scope stack (`beginCloudActivity`) — overlapping save/open/download operations can no longer clear each other's status.
- **Outbox TOCTOU**: enqueue reads and writes in a single IDB readwrite transaction.
- **Reconcile race**: generation-guarded; a stale reconcile can't overwrite fresher UI state. Welcome-seed count is recomputed after reconciliation.
- **Adapter resilience**: `listCanvases` degrades per-canvas when one meta fetch fails; `deleteCanvas` attempts all three objects and rethrows for outbox retry; new `getCanvasMeta` gives the open path a single-canvas lookup instead of listing the whole namespace.
- **Credentials**: clearing disables the dev env-prefill so "clear" actually clears; re-saving re-enables it.
- **Card a11y**: the workspace card is now a `role=button` div with real `<button>` actions inside, visible on keyboard focus via `group-focus-within`.
- **User guide**: entry point corrected to Settings → Cloud storage; security note now states the localStorage same-origin tradeoff explicitly.

Also fixed along the way: the effect-ordering test mock was missing `drawVectorMultiStyleFills` (2 pre-existing test failures in this branch's diff).

rcoenen:
@dannote Heads-up on the red CI: it's not this PR — the repo-wide git-LFS credentials (R2 token `c163e722…` behind the LFS proxy) have expired, so every job dies in setup fetching fixtures. Master's scheduled Heavy-tests run fails identically. Once the token is rotated and `.lfsconfig` updated, I'm happy to rebase/re-run here.

src/components/CodePanel.vue
```diff
@@ -14,6 +13,14 @@ import Tip from '@/components/ui/Tip.vue'
 
 import type { JSXFormat } from '@open-pencil/core/design-jsx'
 
+// Prism language components read the global `Prism` at eval time, and the
+// bundler may order the CJS interop so a component runs before core assigns
+// it (boots the app into "Prism is not defined"). Expose the global from the
+// ESM import and load the jsx grammar dynamically instead.
+;(globalThis as { Prism?: typeof Prism }).Prism = Prism
+const grammarVersion = ref(0)
+void import('prismjs/components/prism-jsx').then(() => (grammarVersion.value += 1))
```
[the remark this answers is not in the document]
rcoenen: Done — added `.catch` so a failed `prism-jsx` chunk load no longer produces an unhandled rejection; highlight still falls back to `Prism.languages.javascript`.

src/app/cloud/registry.ts
```diff
@@ -0,0 +1,27 @@
...
+const S3_PROVIDER: CloudProviderDef = {
+  id: 's3-compatible',
+  label: 'S3 compatible',
+  description: 'AWS S3, Backblaze B2, Cloudflare R2, MinIO, and other S3-compatible buckets'
+}
+
+export const CLOUD_PROVIDERS: CloudProviderDef[] = [S3_PROVIDER]
+
+// S3-compatible is the only provider today; branch on id when a second one lands.
+export function getCloudProviderDef(_id: CloudProviderId): CloudProviderDef {
```
dannote, line 18: This is not yet a real provider registry: both functions ignore `providerId`, and the factory accepts only `S3CompatibleConfig`. Adding Drive, WebDAV, or another backend would still require changing the registry, credential module, settings panel, and call sites.

Please make each provider definition own its typed configuration validation and adapter factory, then resolve a discriminated connection configuration by ID. `CloudStorageAdapter` can remain the backend-neutral operational interface, while the sync engine consumes a concrete connection rather than S3-shaped globals.

src/components/chat/ProviderSettings/CloudStorageSection.vue
```diff
@@ -0,0 +1,409 @@
...
+  buildCorsConfigurationJson,
+  collectCloudCorsOrigins,
+  isLikelyCorsOrNetworkError
+} from '@/app/cloud/s3/cors'
+import type { CloudStorageUsage } from '@/app/cloud/types'
+import { toast } from '@/app/shell/ui'
+import { isTauri } from '@/app/tauri/env'
+import SecretInput from '@/components/Settings/SecretInput.vue'
+import AppSelect from '@/components/ui/AppSelect.vue'
+import ProviderSettingsField from '@/components/chat/ProviderSettings/ProviderSettingsField.vue'
```
dannote, line 37: Cloud settings should not depend on the chat component namespace. This also adds a second secret input and a PascalCase `components/Settings/` domain while the ongoing component refactor is moving toward lowercase domain namespaces and controlled shared primitives.

Please extract provider-setting presentation controls—field, input, secret input, connection status—into a neutral lowercase `settings/` domain or `ui/` when genuinely generic. Compose separate `settings/ai` and `settings/cloud` panels around those controls, and mount one shell-owned Settings dialog rather than importing the full dialog into multiple views. App-specific provider state should remain in the app; only genuinely reusable headless primitives should move into `@open-pencil/vue`.

src/app/cloud/credentials.ts
```diff
@@ -0,0 +1,90 @@
...
+  s3Bucket.value = ''
+  s3AccessKeyId.value = ''
+  s3SecretAccessKey.value = ''
+  // Drop local mirror + outbox so a later account doesn't see orphaned caches.
+  void import('@/app/cloud/sync')
+    .then((m) => m.clearCloudLocalMirror())
+    .catch((error: unknown) => console.warn('[Cloud] failed to clear local mirror:', error))
+}
+
+export function setS3Credentials(config: S3CompatibleConfig) {
```
dannote, line 84: Changing credentials does not change or clear the globally named local database and outbox. Because rows and jobs identify the backend only as `s3-compatible`, switching from bucket A to bucket B can show A’s cached canvases and process A’s pending jobs through B’s adapter.

Please introduce a stable connection/account identity derived from the provider and non-secret endpoint/bucket configuration, and scope local metadata, blobs, and outbox jobs to it. Switching connections should pause sync and commit atomically after validation rather than mutating persisted refs field by field.

This would also be a good boundary for a shared credential store used by both AI and cloud integrations, with a secure desktop backend and explicit browser persistence, instead of adding another provider-specific `localStorage` implementation.

src/app/cloud/s3/client.ts
```diff
@@ -0,0 +1,324 @@
+import { AwsClient } from 'aws4fetch'
```
dannote, line 1: Please replace the custom `aws4fetch` protocol layer with a lazy-loaded AWS SDK v3 implementation using `@aws-sdk/client-s3` and `@aws-sdk/lib-storage`.

I measured the required AWS client, multipart uploader, and browser progress handler at approximately 108 KB gzip—about 98 KB more than the current implementation, but only around 2.5% of the existing generated JavaScript. This is an acceptable tradeoff for typed S3 commands, pagination, structured errors, multipart uploads, checksums, CORS reads/writes, and conditional-write support, provided none of it enters the initial router chunk.

Keep every AWS import behind a dynamic `src/app/cloud/s3/sdk.ts` boundary. Use the official XHR handler in the browser and a custom Smithy request handler backed by `tauriFetch` on desktop. Please also add a build check confirming the SDK remains in a separate lazy chunk with a gzip budget around 115 KB.

src/app/cloud/s3/adapter.ts
```diff
@@ -0,0 +1,216 @@
...
+          )
+        }
+        throw error
+      }
+    },
+
+    async testConnection(): Promise<CloudConnectionTestResult> {
+      // Same automatic step as `aws s3api put-bucket-cors`: configure web origins on the bucket.
+      // Desktop: almost always succeeds. Web: succeeds only if the browser is already allowed.
+      let cors = await ensureWebCorsOnBucket(config)
```
dannote, line 78: `PutBucketCors` is a replacement operation, not an additive one. AWS explicitly documents that an existing bucket CORS configuration is replaced. Calling it unconditionally from Test connection can therefore delete unrelated rules from a bucket that the PR otherwise advertises as safe to share.

Please do not mutate bucket-wide CORS automatically without explicit confirmation. If automatic setup remains, first read the existing configuration and upsert an identifiable OpenPencil rule while preserving every unrelated rule. Add coverage for an existing multi-rule configuration.

AWS reference: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketCors.html

src/app/cloud/s3/client.ts
```diff
@@ -0,0 +1,324 @@
...
+  // Sign with aws4fetch, then re-issue with url+init so the body keeps a known length.
+  // Passing only the signed Request object can drop Content-Length on some runtimes.
+  const signed = await client.sign(url, {
+    ...init,
+    headers,
+    credentials: 'omit'
+  })
+  let res: Response
+  try {
+    if (onUploadProgress && typeof XMLHttpRequest !== 'undefined') {
```
dannote, line 135: `putCanvas()` always supplies a progress callback, so desktop uploads select this raw WebView XHR path whenever `XMLHttpRequest` exists. That bypasses `cloudFetch()` and therefore bypasses `tauriFetch`, which is the transport intended to avoid desktop CORS restrictions.

This path also has no timeout or abort handling, so one stalled upload can leave the single-flight sync pump blocked indefinitely. Please choose the transport by platform, preserve the Tauri HTTP bridge on desktop, and give the browser progress transport the same timeout and cancellation semantics. This needs tests for both transport selections.

src/app/cloud/s3/adapter.ts
```diff
@@ -0,0 +1,216 @@
...
+      return canvases
+    },
+
+    async getCanvas(id: string, onProgress) {
+      const bytes = await getObject(config, canvasFigKey(id), onProgress)
+      if (!bytes) throw new Error(`Canvas not found: ${id}`)
+      return bytes
+    },
+
+    async putCanvas(id: string, bytes: Uint8Array, meta: CloudCanvasMeta, onProgress) {
```
dannote, line 162: The revision check in the outbox protects only against stale writes from the same local database. There is no remote ETag/version precondition here, so device A can edit offline, device B can upload a newer revision, and A will silently overwrite B when it reconnects. The declared `conflict` state is never reached.

Please carry a remote version token through the adapter and local metadata, and use a conditional write such as `If-Match` where supported. A failed precondition should preserve both versions or surface a conflict rather than retrying as an ordinary upload. If a target provider cannot conditionally overwrite objects, use an equivalent immutable-revision/manifest strategy. Add a two-device stale-write regression test.

dannote (changes requested):
Thanks for taking on a substantial piece of infrastructure. I left comments around provider boundaries, settings composition, S3 transport, and sync safety.

Before this lands, I think the cloud work needs to be separated from the unrelated renderer, page-color, and Prism changes, then rebased onto master. The targeted tests pass, but `format:check` rewrites three files, `check:vue` fails on the `prism-jsx` import, and no test currently exercises the sync engine, reconciliation, account switching, or remote-conflict paths.

dannote:
Thanks, Rob! This work has been adapted and integrated on `master`, with your co-author credit preserved. The original branch is now superseded, so I’m closing this PR. It will ship in v0.14.0.

PR #392, opened 2026-07-11
fix(canvas): import dropped SVG files as vector nodes

rcoenen:
## Summary

Dragging an SVG file onto the canvas navigated the browser to the file (opened it in a new tab) instead of importing it. The canvas drop handler only accepted raster image types, so the `dragover` was never claimed for SVGs and the browser's default drag behavior won.

SVG drops now import as editable vectors: a frame of VECTOR nodes at the drop position, parsed by the same code path as the existing `import_svg` tool, selected after import, and undoable as "Import SVG".

## What changed

- `packages/core/src/tools/create/svg.ts` — extracted `createSvgNodes(graph, parentId, svg, options)` from the `import_svg` tool so the parser is callable outside the tool adapter; the tool now delegates to it
- `packages/core/src/editor/clipboard/svg.ts` — new `placeSvgFiles(files, cx, cy)` clipboard action mirroring `placeImageFiles`: row layout around the drop point, paste-target parent, subtree undo via the paste helpers
- `packages/vue/src/canvas/drop/use.ts` — the canvas drop accepts `image/svg+xml` (with a `.svg` filename fallback for drags that lose the MIME type) alongside rasters and routes each kind to its handler

## Validation

- [x] `bun test tests/engine/editor/svg-drop-import.test.ts` — frame sizing from viewBox/width/height, per-path fills and strokes, null on unsupported markup
- [x] `bun test tests/engine/tools/vectorize.test.ts` (neighbouring SVG consumers unaffected)
- [x] Typecheck + lint clean
- [x] Manual: dropping `.svg` imports vectors at the cursor; dropping `.png` still places an image; undo removes the imported frame

## Summary by CodeRabbit

* **New Features**
  * Added support for dropping SVG files onto the canvas, converting them into editable vector frames positioned at the drop location.
  * Added SVG clipboard placement actions to insert SVG content.
* **Bug Fixes**
  * Improved SVG import to correctly map `viewBox` into node space and preserve per-path fills, strokes, and sizing.
* **Tests**
  * Added automated coverage for SVG node creation, including styling/geometry validation and `null` when no supported SVG elements are present.

### DIFF
--- a/packages/core/src/editor/clipboard/svg.ts
+++ b/packages/core/src/editor/clipboard/svg.ts
@@ -1,0 +1,58 @@
+import { resolvePasteTarget } from '#core/editor/clipboard/paste-target'
+import type { EditorContext } from '#core/editor/types'
+import { createSvgNodes } from '#core/tools/create/svg'
+
+import { deleteIds, recreateSnapshots } from './history'
+import { collectSubtrees } from './subtree-history'
+
+const SVG_GAP = 20
+
+export function createClipboardSvgActions(ctx: EditorContext) {
+  /** Import dropped/pasted SVG files as vector frames centered around (cx, cy). */
+  async function placeSvgFiles(files: File[], cx: number, cy: number) {
+    const parentId = resolvePasteTarget(ctx)
+    const prevSelection = new Set(ctx.state.selectedIds)
+
+    const frames = []
+    for (const file of files) {
+      const markup = await file.text()
+      const frame = createSvgNodes(ctx.graph, parentId, markup, {
+        name: file.name.replace(/\.svg$/i, '') || 'SVG'
+      })
+      if (frame) frames.push(frame)
+    }
+    if (!frames.length) return
+
+    let totalW = 0
+    for (const f of frames) totalW += f.width
+    totalW += SVG_GAP * (frames.length - 1)
+    const maxH = Math.max(...frames.map((f) => f.height))
+
+    let curX = cx - totalW / 2
+    const topY = cy - maxH / 2
+    for (const frame of frames) {
+      ctx.graph.updateNode(frame.id, { x: curX, y: topY })
+      curX += frame.width + SVG_GAP
+    }
+
+    const created = frames.map((f) => f.id)
+    const allNodes = collectSubtrees(ctx.graph, created)
+    const pageId = ctx.state.currentPageId
+    ctx.undo.push({
+      label: 'Import SVG',
+      forward: () => {
+        recreateSnapshots(ctx, allNodes, pageId)
+        ctx.setSelectedIds(new Set(created))
+      },
+      inverse: () => {
+        deleteIds(ctx, created)
+        ctx.setSelectedIds(prevSelection)
+      }
+    })
+
+    ctx.setSelectedIds(new Set(created))
+    ctx.requestRender()
+  }
+
+  return { placeSvgFiles }
+}
--- a/packages/core/src/editor/clipboard/svg.ts
+++ b/packages/core/src/editor/clipboard/svg.ts
@@ -1,5 +1,8 @@
+import type { SceneNode } from '@open-pencil/scene-graph'
+
 import { resolvePasteTarget } from '#core/editor/clipboard/paste-target'
 import type { EditorContext } from '#core/editor/types'
+import { computeAllLayouts } from '#core/layout'
 import { createSvgNodes } from '#core/tools/create/svg'
 
 import { deleteIds, recreateSnapshots } from './history'
@@ -13,7 +16,7 @@ export function createClipboardSvgActions(ctx: EditorContext) {
     const parentId = resolvePasteTarget(ctx)
     const prevSelection = new Set(ctx.state.selectedIds)
 
-    const frames = []
+    const frames: SceneNode[] = []
     for (const file of files) {
       const markup = await file.text()
       const frame = createSvgNodes(ctx.graph, parentId, markup, {
@@ -42,14 +45,17 @@ export function createClipboardSvgActions(ctx: EditorContext) {
       label: 'Import SVG',
       forward: () => {
         recreateSnapshots(ctx, allNodes, pageId)
+        computeAllLayouts(ctx.graph, pageId)
         ctx.setSelectedIds(new Set(created))
       },
       inverse: () => {
         deleteIds(ctx, created)
+        computeAllLayouts(ctx.graph, pageId)
         ctx.setSelectedIds(prevSelection)
       }
     })
 
+    computeAllLayouts(ctx.graph, pageId)
     ctx.setSelectedIds(new Set(created))
     ctx.requestRender()
   }
--- a/packages/core/src/tools/create/svg.ts
+++ b/packages/core/src/tools/create/svg.ts
@@ -1,3 +1,4 @@
+import type { SceneGraph } from '@open-pencil/scene-graph'
 import type { Rect } from '@open-pencil/scene-graph/primitives'
 
 import { parseColor } from '#core/color'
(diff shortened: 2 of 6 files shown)


packages/core/src/editor/clipboard/svg.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const frames = []
+    for (const file of files) {
+      const markup = await file.text()
+      const frame = createSvgNodes(ctx.graph, parentId, markup, {
+        name: file.name.replace(/\.svg$/i, '') || 'SVG'
+      })
+      if (frame) frames.push(frame)
+    }
+    if (!frames.length) return
```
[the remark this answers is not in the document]
rcoenen: Done in 365c8c79 — `const frames: SceneNode[] = []`.

packages/core/src/editor/clipboard/svg.ts
```diff
@@ -0,0 +1,58 @@
...
+        ctx.setSelectedIds(new Set(created))
+      },
+      inverse: () => {
+        deleteIds(ctx, created)
+        ctx.setSelectedIds(prevSelection)
+      }
+    })
+
+    ctx.setSelectedIds(new Set(created))
+    ctx.requestRender()
```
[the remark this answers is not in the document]
rcoenen: Done in 365c8c79 — `computeAllLayouts(ctx.graph, pageId)` now runs after the import and in both undo callbacks, mirroring `pushPasteUndo`.

packages/core/src/tools/create/svg.ts
```diff
@@ -31,50 +33,124 @@ function parseSvgSize(svg: string): { width: number; height: number } {
...
+/** Parse SVG markup into a group of vector nodes (group resize scales the
+ *  vectors). Returns the group root, or null when the markup contains no
+ *  supported elements. */
+export function createSvgNodes(
+  graph: SceneGraph,
+  parentId: string,
+  svg: string,
+  options: CreateSvgNodesOptions = {}
+) {
+  const paths = extractPaths(svg)
```
dannote, line 132: `extractPaths()` is a regex-based Iconify helper, not a general SVG importer. It does not build an SVG tree or correctly handle nested transforms, scoped inheritance, CSS, `<use>`, or gradients. For example, `<g transform="translate(40 30)">…</g>` currently imports its child at `(0, 0)`.

Please move this into a shared `io/formats/svg` import pipeline backed by a real XML/tree parser. Both `import_svg` and canvas drop should use that pipeline rather than making the editor depend on a ToolDef module.

packages/core/src/tools/create/svg.ts
```diff
@@ -31,50 +33,124 @@ function parseSvgSize(svg: string): { width: number; height: number } {
...
 
+/** Map a parsed network from viewBox units into node-space pixels. */
+function fitNetworkToNode(
+  network: ReturnType<typeof parseSVGPath>,
+  viewBox: Rect | null,
+  width: number,
+  height: number
+) {
+  if (!viewBox || viewBox.width <= 0 || viewBox.height <= 0) return network
+  const sx = width / viewBox.width
```
dannote, line 44: Could we use the existing `svgpath` dependency instead of adding another path transformation implementation? It already supports composing matrix, translation, scale, rotation, and skew transforms before passing the normalized path to `parseSVGPath()`.

The viewport matrix also needs to honor `preserveAspectRatio`. For example, a `viewBox="0 0 100 50"` inside `width="200" height="200"` should remain `200×100` and be centered vertically; this code stretches it to `200×200`. Please compute the SVG viewport and ancestor/element transform matrix once, then apply it through `svgpath`.

packages/vue/src/canvas/drop/use.ts
```diff
@@ -40,7 +46,8 @@ export function useCanvasDrop(canvasRef: Ref<HTMLCanvasElement | null>, editor:
     const sy = e.clientY - rect.top
     const { x: cx, y: cy } = editor.screenToCanvas(sx, sy)
 
-    void editor.placeImageFiles(files, cx, cy)
+    if (images.length) void editor.placeImageFiles(images, cx, cy)
```
dannote, line 49: A mixed raster + SVG drop starts two independent asynchronous placement transactions at the same center. The batches overlap, their undo entries are ordered by whichever promise finishes first, and whichever finishes last replaces the selection.

Could all accepted assets go through one preparation and placement transaction so the complete dropped file list is laid out, selected, and undone together?

packages/vue/src/canvas/drop/use.ts
```diff
@@ -49,7 +56,8 @@ export function useCanvasDrop(canvasRef: Ref<HTMLCanvasElement | null>, editor:
 function hasImageFiles(e: DragEvent): boolean {
   if (!e.dataTransfer?.types.includes('Files')) return false
   for (const item of e.dataTransfer.items) {
-    if (item.kind === 'file' && ACCEPTED_TYPES.has(item.type)) return true
+    if (item.kind !== 'file') continue
+    if (ACCEPTED_TYPES.has(item.type) || item.type === SVG_TYPE) return true
```
dannote, line 60: The filename fallback is only applied later by `filterSvgFiles()`. During `dragover`, this still rejects an SVG whose `DataTransferItem.type` is empty. Because `dragover` is not cancelled, the target might never accept the drop and the browser can still navigate to the file.

Please inspect `item.getAsFile()?.name` here, or claim file drags first and validate their types on drop.

packages/core/src/editor/clipboard/svg.ts
```diff
@@ -0,0 +1,64 @@
...
+    const topY = cy - maxH / 2
+    for (const frame of frames) {
+      ctx.graph.updateNode(frame.id, { x: curX, y: topY })
+      curX += frame.width + SVG_GAP
+    }
+
+    const created = frames.map((f) => f.id)
+    const allNodes = collectSubtrees(ctx.graph, created)
+    const pageId = ctx.state.currentPageId
+    ctx.undo.push({
```
dannote, line 44: This duplicates `pushPasteUndo()` from `editor/clipboard.ts`, and `bun run test:dupes` reports this exact clone. The same gate also reports duplicated path styling between `tools/create/svg.ts` and `icons/render.ts`.

Please extract and reuse the shared undo operation with a caller-supplied label, and share the existing path-style application rather than maintaining parallel implementations.

tests/engine/editor/svg-drop-import.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { createSvgNodes } from '#core/tools/create/svg'
+
+import { expectDefined } from '#tests/helpers/assert'
+
+const STAR_SVG = `<svg viewBox="0 0 24 24" width="48" height="48">
+  <path d="M12 2 L15 9 L22 9 L16 14 L18 21 L12 17 L6 21 L8 14 L2 9 L9 9 Z" fill="#ff6600"/>
+  <path d="M0 0 H24 V24 H0 Z" fill="none" stroke="#333333" stroke-width="2"/>
+</svg>`
+
+describe('SVG drop import', () => {
```
dannote, line 14: These tests only call `createSvgNodes()`, so they do not cover the behavior introduced by this PR: file reading, placement, selection, undo/redo, MIME fallback, or mixed drops.

The existing `tests/engine/io/svg/import.test.ts` also now fails because `import_svg` returns a `GROUP` instead of its documented and tested `FRAME`. Please either preserve that contract or update it intentionally, and add coverage for the actual editor/drop path.

dannote (changes requested):
Thanks for taking this on. I left a few comments around the SVG import boundary, viewport semantics, mixed-file placement, and coverage. The branch also currently regresses the existing SVG tool test and fails the duplication gate.

dannote:
Thanks, Rob! This has been adapted and landed on `master` as `7e8feb99`, with your co-author credit preserved. I’m closing the superseded PR.

PR #402, opened 2026-07-16
fix(clipboard): paste image between two files

yoohoo360:
Summary
Fixes a clipboard paste issue where copying from different source files (especially unfinished Figma assets) into Open Pencil fails to produce an image.

This change improves cross-source paste handling by accounting for cases where Figma does not place a direct image stream in the clipboard.
When copying between Figma files, data transfer may rely on Figma’s internal object storage mechanism instead of standard clipboard image data, which caused Open Pencil paste to fail.

What changed
Improved paste logic for cross-file clipboard content.
Added handling/fallback for clipboard payloads that do not contain a direct image stream.
Fixed a scenario where unfinished or in-progress Figma content could not be pasted into Open Pencil.
Added clearer detection for Figma-originated clipboard data patterns.

### Validation

- [x] `bun run check`
- [ ] Tests added or updated, or not needed because: explain here
- [ ] CHANGELOG.md updated, or not needed because: explain here

## Summary by CodeRabbit

* **New Features**
  * Added undo and redo support for images pasted from Figma, ensuring pasted image content is restored correctly.
  * Exported nodes with image fills now include their associated image content.
* **Bug Fixes**
  * Improved clipboard history handling so undoing and redoing image pastes no longer loses image data.

### DIFF
--- a/packages/core/src/editor/clipboard.ts
+++ b/packages/core/src/editor/clipboard.ts
@@ -1,4 +1,4 @@
-import type { SceneNode } from '@open-pencil/scene-graph'
+import { computeImageHash, type SceneNode } from '@open-pencil/scene-graph'
 import type { Vector } from '@open-pencil/scene-graph/primitives'
 
 import {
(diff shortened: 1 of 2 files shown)


dannote:
Thanks for investigating this. The current approach treats every Kiwi blob as image data, even though the blob array also contains vector geometry and text payloads. It also stores image bytes under a different hash from the one referenced by Figma image fills, so real Figma images remain unresolved while non-image blobs are added to `graph.images`.

Because the implementation needs a different approach rather than incremental changes, I’m going to close this PR. A follow-up would be welcome with a real clipboard fixture from the failing case and a regression test showing that the pasted fill’s `imageHash` resolves in `graph.images`. The existing explicit image map in the OpenPencil clipboard format may be a better foundation.

PR #410, opened 2026-07-17
feat(keyboard): add Figma-style opacity shortcuts 1-9/0

sadkodev:
Add opacity keyboard shortcuts matching Figma: pressing Digit1-Digit9 sets the selected layers to 10%-90% opacity, and Digit0 sets 100%.

- Add 'selection.setOpacity' EditorCommandId with registry metadata (shortcut display '1-9, 0') so it appears in the command cheatsheet
- Implement the command in selection.ts using editor.undo.runBatch + updateNodeWithUndo, following the toggleVisibility/setBlendMode pattern
- Add setOpacityTarget closure in useEditorCommands so the external keyboard handler can inject the digit value before runCommand
- Add canSetOpacity capability (enabled when there is a selection)
- Add bindOpacityKeys handler in src/app/shell/keyboard/opacity.ts, following the nudging.ts external-handler pattern since the opacity value comes from event.code and cannot be expressed as a fixed keybinding in the registry
- Guard against text editing, number field focus, and modifier keys
- Add i18n label 'setOpacity' across all 8 locales
- Add unit tests for single/multi selection, undo batch, and no-op

> Security vulnerability? Do not open a public PR. Report it privately through GitHub Security Advisories: https://github.com/open-pencil/open-pencil/security/advisories/new

Before opening a PR, read `CONTRIBUTING.md` and `AGENTS.md`. PRs should explain the intent, meaningful changes, and validation. Placeholder, non-English, unrelated, or otherwise unreviewable PRs may be closed by maintainers.

### What changed

- Add a concise summary here.

### Validation

- [ ] `bun run check`
- [ ] Tests added or updated, or not needed because: explain here
- [ ] CHANGELOG.md updated, or not needed because: explain here

## Summary by CodeRabbit

* **New Features**
  * Set opacity for the current selection using `1`–`9` and `0` (top-row and numpad), following Figma-style shortcuts.
  * Support multi-digit opacity entry (e.g., `28` → `28%`, `00` → `0%`, `100` → `100%`) with digit buffering and reset on inactivity.
  * Opacity updates apply to multi-selections and are grouped into a single undoable action; values are constrained to `0%–100%` and clamped.
* **Documentation**
  * Added localized “Set opacity” command labels across supported languages.


src/app/shell/keyboard/opacity.ts
```diff
@@ -0,0 +1,52 @@
...
+  Numpad6: 0.6,
+  Digit7: 0.7,
+  Numpad7: 0.7,
+  Digit8: 0.8,
+  Numpad8: 0.8,
+  Digit9: 0.9,
+  Numpad9: 0.9
+} as const
+
+export function bindOpacityKeys({ store, setOpacityTarget, runCommand }: OpacityShortcutOptions) {
```
dannote: Could we route this through the existing `registerKeyboardShortcuts`/`tinykeys` path instead of adding another window listener? `tinykeys` supports digit regex bindings like `([0-9])`, and the registry already owns the common input and focus guards. This would also avoid the NumLock-off numpad issue, since `event.key` is no longer numeric.

src/app/shell/keyboard/opacity.ts
```diff
@@ -0,0 +1,52 @@
...
+  useEventListener(window, 'keydown', (e: KeyboardEvent) => {
+    if (isEditing(e)) return
+    if (store.state.editingTextId) return
+    if (store.state.numberFieldFocused) return
+    if (e.metaKey || e.ctrlKey || e.altKey || e.shiftKey) return
+    if (store.state.selectedIds.size === 0) return
+
+    if (!(e.code in DIGIT_OPACITY)) return
```
dannote: This handles each digit independently, but Figma also combines rapidly entered digits: `28` sets 28% and `00` sets 0%. Here those inputs end at 80% and 100%. Could we add the short digit buffer and cover both single- and two-digit behavior? Figma's forum also confirms the `00` behavior: https://forum.figma.com/archive-21/keyboard-shortcut-for-0-opacity-36494

tests/engine/editor/opacity.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      x: 0,
+      y: 0,
+      width: 50,
+      height: 50
+    })
+    editor.select([rect.id])
+    return { editor, rect }
+  }
+
+  function setOpacity(editor: ReturnType<typeof createEditor>, opacity: number) {
```
dannote: These tests copy the production opacity logic into a local helper, so they would still pass if the command or keyboard wiring broke. Could we extract the mutation into a real core editor action and test that directly, then add a small E2E test for the keyboard behavior?

dannote (changes requested):
Thanks for working on this! I left a few comments for you to take a look at.

src/app/shell/keyboard/opacity.ts
sadkodev: Thanks for the suggestion! I’ll dig into the existing `registerKeyboardShortcuts` / `tinykeys` registry and rework the implementation to route through it instead of adding another window listener. I’ll also take a closer look at the regex digit binding `(([0-9]))` and the existing `input/focus` guards to handle the `NumLock-off` case properly. Appreciate the catch!

src/app/shell/keyboard/opacity.ts
sadkodev: Good point, thanks! I had been thinking only in terms of single digits and completely forgot that the input also accepts two-digit values as percentages `(like 28 = 28% or 00 = 0%)`. I’ll keep that in mind and rework the handler to include a short digit buffer so it supports both single and multi-digit input. Good catch!

tests/engine/editor/opacity.test.ts
sadkodev: Exactly, you’re right. The local helper is just mirroring the production logic, so the tests would pass even if the real implementation broke. I’ll improve my testing strategy by extracting the mutation into a real core editor action, unit-testing that directly, and adding a small E2E test to cover the keyboard wiring. Thanks for pushing me on this!

dannote (approved):
Thanks for addressing the review. I also tightened the final edge cases: NumLock-off navigation is no longer intercepted, buffered digits now coalesce into one undo step, the unrelated dependency was removed, and keyboard behavior has end-to-end coverage. The focused tests and full quality gate pass on the rebased branch.

sadkodev:
Nice, all the edge cases are covered and the rebase looks clean. LGTM.

PR #416, opened 2026-07-21
feat(export): add editable PPTX export

greekr4:
Hi! It's been a while — my last PRs here were the dash-pattern stroke fixes (#237/#238) and the color-picker undo batching (#187) back in the spring, and it's been nice to see some of those ideas live on in your later refactors. I've kept building on OpenPencil in the meantime, and this time I'm back with something a bit bigger that's been genuinely useful for us in production: exporting decks as *editable* PowerPoint files. Hope it's useful upstream too!

### Summary

Adds editable PowerPoint (`.pptx`) export as a new io format adapter, wired into the app's **File → Export selection** menu, the CLI file mode and the SDK. Unlike a rasterized deck, text, rectangles, ellipses and lines are converted to **native PowerPoint elements** — the exported deck stays editable in PowerPoint, Keynote and LibreOffice (retype text, move shapes, restyle fills). Content that PPTX can't express natively (gradients, multiple fills, blend modes, blurs, masks, vector/icon artwork) falls back to transparent PNG images rendered through the existing raster pipeline, so the deck looks identical either way.

Each selected top-level frame becomes one slide; slide dimensions derive from the first frame's aspect ratio. Unit conversion uses exact formulas (1in = 96px = 72pt) for positions, font sizes, letter spacing, line height and stroke widths — no hand-tuned factors.

`pptxgenjs` is added as a core dependency but is **loaded lazily at export time only** (mirroring the `jspdf` pattern), so it never enters the initial bundle.

### What changed

- `packages/core/src/io/formats/pptx/export.ts` — new `renderNodesToPPTX(graph, pageId, nodeIds, options)`:
  - recursive scene-graph walk emitting native text (`styleRuns` merged into per-run PPT styling: font, size, bold, italic, color, letter spacing, underline/strikethrough), native shapes (rect/roundRect/ellipse/line with fills, strokes, corner radii, rotation, shadows) and native pictures for image-fill leaves
  - PNG fallback for non-representable nodes; icon-style containers (all visible children are vectors) rasterize as **one** composed image instead of per-path fragments
  - fallback rasterization renders the subtree **in isolation** (ancestor paints cleared in the extracted graph) so images keep transparency instead of baking in the slide background; uses the headless raster pipeline, or the live renderer when an `IOContext` provides one
- `packages/core/src/io/formats.ts` — `pptxFormat` adapter (`category: 'print'`, all export scopes) registered in `BUILTIN_IO_FORMATS`.
- `src/app/shell/menu/schema.ts`, `app-menu.ts`, `src/app/document/export/create.ts` — PPTX entry in File → Export selection (next to PNG/SVG/.fig; label is a format acronym, so no i18n keys needed).
- `packages/cli/src/commands/export.ts` — `--format pptx` in file mode (`openpencil export file.fig --format pptx`); app mode reports file-mode-only for now.
- `tests/engine/io/pptx/export/render.test.ts` — 6 unit tests that unzip the produced archive and assert slide count, native text runs (partial bold survives as its own run), native shape colors, gradient→media fallback with conversion stats, single-image icon rasterization, and slide background mapping — using a stubbed rasterizer so tests stay CanvasKit-free.
- `CHANGELOG.md` — Unreleased → Added entry.

### Screenshots
pptx-pr-1-select-frames
pptx-pr-2-export-menu
pptx-pr-3-powerpoint-editing

### Validation

- [x] `bun run check`
- [x] Tests added — `tests/engine/io/pptx/export/render.test.ts` (6 tests). Full `bun run test:unit` matches the master baseline on this machine (the heavy `.fig` fixture suites are flaky on clean `master` too, so nothing new there).
- [x] CHANGELOG.md updated
- Real-world check: exported `tests/fixtures/gold-preview.fig` and a two-slide demo document via CLI; decks open in LibreOffice and PowerPoint with editable text/shapes and transparent icon images (screenshots in a comment below).

Possible follow-ups if there is interest, kept out of this PR to stay reviewable: font embedding into the OOXML archive (worth a licensing discussion first) and PPTX in the app-connected CLI mode.

## Summary by CodeRabbit

* **New Features**
  * Added editable PowerPoint (`.pptx`) export for selections, pages, and documents via the File menu, app menu, CLI, and SDK.
  * Exported content keeps basic shapes/text/lines editable; advanced effects (for example gradients, masks, blends, vectors, icons) fall back to embedded images when needed.
  * Whole-document exports include all pages by default; app-mode `.pptx` export is constrained (page-only behavior).
* **Documentation**
  * Updated the changelog with the editable-vs-fallback behavior.
* **Tests**
  * Added automated coverage for PPTX slide output, fallbacks (with stats), clipping/masking behavior, and multi-page export.


packages/core/src/io/formats/pptx/export.ts
```diff
@@ -0,0 +1,580 @@
...
+  opacity: number,
+  reason: string | null
+): Promise<void> {
+  const data = await ctx.rasterize([node.id], ctx.fallbackScale)
+  if (!data) {
+    ctx.stats.skipped += 1
+    return
+  }
+  const box = nodeBox(ctx, node)
+  ctx.slide.addImage({
```
dannote, line 516: Rotated nodes can end up misplaced here. `getAbsolutePosition()` already includes parent transforms, while PowerPoint applies `node.rotation` again around the element’s center. Nested rotation is also lost.

Could we derive both position and rotation from the world transform, or rasterize transforms that PPTX can’t represent? A test with a child inside a rotated frame would help.

packages/core/src/io/formats/pptx/export.ts
```diff
@@ -0,0 +1,580 @@
...
+    }
+  }
+}
+
+/** Why a node cannot be converted natively. Null means native conversion is possible. */
+function getFallbackReason(graph: SceneGraph, node: SceneNode): string | null {
+  if (!SIMPLE_BLENDS.has(node.blendMode)) return 'blend mode'
+  if (node.effects.some((e) => e.visible && e.type !== 'DROP_SHADOW' && e.type !== 'INNER_SHADOW'))
+    return 'blur effect'
+  if (node.childIds.some((id) => graph.getNode(id)?.isMask)) return 'contains mask'
```
dannote, line 288: We should account for `clipsContent` here. PPTX receives the frame and its children as separate elements, so children outside a clipped frame become visible.

Could clipped containers fall back to a single image? Please add a test with a child extending beyond the frame.

packages/core/src/io/formats.ts
```diff
@@ -286,6 +287,41 @@ export const pdfFormat: IOFormatAdapter = {
...
 
+export const pptxFormat: IOFormatAdapter = {
+  id: 'pptx',
+  label: 'PowerPoint',
+  role: 'derived-export',
+  category: 'print',
+  extensions: ['pptx'],
+  mimeTypes: ['application/vnd.openxmlformats-officedocument.presentationml.presentation'],
+  support: {
+    exportDocument: true,
```
dannote, line 311: `exportDocument` currently exports only the first page because the shared resolver selects `graph.getPages()[0]`.

Could PPTX document export collect frames from every page, or disable document support for now? A two-page test would make the expected behavior clear.

dannote (dismissed):
Thanks for working on this. PPTX export will be a useful addition.

I found three issues to address before merging: rotated-node placement, clipped-frame content, and document export stopping after the first page. Please also take care of the existing comments about frame styling and line transparency.

Once these issues are fixed and covered by tests, I would be happy to merge this.

greekr4:
I’ll take a look!

greekr4:
All five points fixed in 5ae21a8b, each with a test. Screenshots are a fixture deck rendered by LibreOffice, before vs after.

**Rotation** — placement now comes from the node's transform relative to the slide frame, not `getAbsolutePosition()` + `node.rotation`. Ancestor rotation composes (B = 20° + 25° = 45°), mirrored basis becomes `flipH`. Every shape should sit centred on its dot:

1-rotation-before
1-rotation-after

**Clipping** — a clipping frame rasterises when a child actually overflows; if everything fits it stays native. F must stay cropped, G is the control, H fits:

2-clipping-before
2-clipping-after

**Document export** — now spans all pages. Slide 4 is on page 2; it used to stop at three.

4-second-page-after

**Frame styling** — stroke, radius and translucency map to a native slide-sized shape; plain solid fill still becomes the slide background, now with `fill.opacity * color.a`. Only gradients/images/blurs/blend modes rasterise, and *without* the children — baking them in drew everything twice (see the before shot):

3-frame-paint-before
3-frame-paint-after

**Line alpha** — `LINE` now multiplies by the paint's `opacity` and `color.a`.

---

14 tests in `render.test.ts`, each verified to fail on the previous commit. Style helpers split into `pptx/style.ts` for `max-lines`. `bun run check` passes.

Skipped the batching nitpick — fallbacks render in isolation to keep transparency, batching would tangle that for a small win.

Unrelated, hit while building the fixture: `exportFigFile()` shifts rotated nodes on write (`(100, 470)` at `-10°` reads back as `(110.4, 498.4)`, stable after). Not from this PR — the canvas shows the same shift. Can open a separate issue if useful.

PR #433, opened 2026-07-28
fix(canvas): keep canvas alive when toggling the UI

### LINKED ISSUE #432: Canvas goes blank with "Cannot set properties of null (setting 'be')" when toggling the UI
### What happened?

Hiding the UI (the **Toggle UI** button, or `MOD+\`) blanks the canvas and raises an uncaught `TypeError: Cannot set properties of null (setting 'be')` from CanvasKit. Error toasts appear and the canvas stops rendering entirely — the page background and all layers disappear, while selection overlays keep drawing.

The canvas does not recover until the UI is toggled back on.

### Platform

Web (browser)

### Steps to reproduce

1. Open the editor on a wide window (see note on width below).
2. Draw a rectangle so there is something on the canvas.
3. Click **Toggle UI** (top-right), or press `MOD+\`.
4. Error toast appears; the scene layer goes blank.

**Reproduces most reliably on a wide viewport with a HiDPI display.** On my machine the canvas area is 2998×1490 CSS px at `devicePixelRatio` 2. With the side panels open the viewport is ~1919 px wide and everything is fine; hiding the UI widens it to ~2998 px and it breaks immediately — which is exactly the trigger condition (see below).

### Affected area

Canvas / rendering

### Version or build

`master` @ `679ee9f9`, app version 0.13.2

### Console output or logs

```
Uncaught TypeError: Cannot set properties of null (setting 'be')
    at a.Surface.makeSurface (canvaskit-wasm.js)
    at createSceneBackingSurface (packages/core/src/canvas/renderer/retained-backing.ts:178)
    at recordSceneBacking (packages/core/src/canvas/renderer/retained-backing.ts:279)
    at renderSceneBacking (packages/core/src/canvas/renderer/retained-backing.ts:302)
    at render (packages/core/src/canvas/renderer/pipeline.ts:105)
    at renderFromEditorState (packages/core/src/canvas/renderer.ts:204)
    at renderNow (packages/vue/src/canvas/surface/lifecycle.ts:50)
    at init (packages/vue/src/canvas/surface/kit-loader.ts:16)          <- variant A
    at resizeCanvas (packages/vue/src/canvas/surface/lifecycle.ts:75)   <- variant B
```

### Investigation

There appear to be two independent causes; fixing only one leaves the bug reproducible.

**1. The canvas is unmounted and remounted on every UI toggle.**

`src/views/EditorView.vue` writes `<EditorCanvas />` once inside each of four `v-if` / `v-else-if` / `v-else` layout branches (desktop / mobile / collapsed / bare). Toggling `showUI` switches branches, so Vue destroys the whole subtree and builds a new one. Node reuse is impossible here because the branch roots differ (`SplitterGroup` vs plain `<div>`) and each branch carries a distinct `:key` prefix (`activeTab?.id`, `'mobile-' + …`, `'collapsed-' + …`, `'bare-' + …`).

The remount tears down the WebGL surface and the Skia renderer. Because `kit-loader.ts` re-awaits `getCanvasKit()`, waits a `requestAnimationFrame`, and reloads fonts before the first render, the newly mounted `<canvas>` sits at the browser default size in the meantime. Inspecting the DOM right after a toggle shows exactly that:

```js
[{cw: 2998, ch: 1531, w: 300, h: 150, ready: undefined},
 {cw: 2998, ch: 1531, w: 300, h: 150, ready: undefined}]
```

CSS size is correct, but the backbuffer is still `300×150` and `data-ready` was never set.

**2. The retained scene backing exceeds `MAX_TEXTURE_SIZE`.**

`sceneBackingGeometry` in `packages/core/src/canvas/renderer/retained-backing.ts` multiplies the viewport by `SCENE_BACKING_SCALE` (3), and `createSceneBackingSurface` then multiplies by `dpr` — with no upper bound. At 2998×1490 CSS px and `dpr` 2:

```
ceil(2998 * 3) * 2 = 17988 px wide
gl.getParameter(gl.MAX_TEXTURE_SIZE) = 16384
```

The allocation is impossible, so it fails. With the panels open the same math gives `1919 * 3 * 2 = 11514`, comfortably under the limit — which is precisely why this only shows up once the panels close and the viewport widens.

`createSceneBackingSurface` is already typed `Surface | null` and its call site guards with `if (!surface) return`, but CanvasKit **throws** instead of returning `null`, so that guard never runs and the error escapes as an uncaught `TypeError`.

### Before submitting

- [x] Searched existing issues
- [x] Reproduced on the latest `master`

I have a fix for both causes and will open a PR referencing this issue.

Dae-Jeong:
Fixes #432

### Summary

Hiding the UI (**Toggle UI** / `MOD+\`) blanked the canvas and raised an uncaught `TypeError: Cannot set properties of null (setting 'be')` from CanvasKit. Two independent causes were involved — fixing either one alone still left the bug reproducible, so both are addressed here.

### What changed

**1. The canvas is no longer remounted when the layout changes.**

`EditorView.vue` wrote `<EditorCanvas />` once inside each of four `v-if` / `v-else-if` / `v-else` layout branches. Toggling `showUI` swapped branches, and Vue destroyed the entire subtree — taking the WebGL surface and the Skia renderer with it. Reuse was impossible because the branch roots differ (`SplitterGroup` vs `<div>`) and each branch had a distinct `:key` prefix.

The canvas is now mounted exactly once, outside every branch, and `<Teleport>`ed into whichever slot the active layout exposes. Teleport moves the existing DOM node rather than recreating it, so the GL context survives. The layout branches themselves became `v-show` so both teleport targets stay in the DOM.

Verified by tagging the `<canvas>` nodes and toggling: the *same* node objects survive (`sameNodes: true`), `data-ready` stays set, and the backbuffer resizes (3838 → 5996 → 3838) instead of collapsing to the `300×150` default.

**2. The retained scene backing is clamped to the GPU texture limit.**

`sceneBackingGeometry` multiplied the viewport by `SCENE_BACKING_SCALE` (3) and `createSceneBackingSurface` multiplied again by `dpr`, unbounded. At 2998×1490 CSS px with `dpr` 2 that requests a **17988 px** wide texture against a `MAX_TEXTURE_SIZE` of **16384**, so the allocation fails. With the panels open the same math yields 11514 px and stays under the limit — which is why the bug only appeared once the panels closed and the viewport widened.

The backing is now clamped so the final pixel size stays within `MAX_TEXTURE_SIZE`, read once from the GL context in the `SkiaRenderer` constructor (falling back to the WebGL2-mandated floor of 2048 when unavailable). A smaller backing only costs extra re-records while panning; `worldWidth`/`worldHeight` derive from the clamped values, so the existing coverage checks stay consistent.

Separately, `createSceneBackingSurface` now catches. It was already typed `Surface | null` with an `if (!surface) return` guard at the call site, but CanvasKit *throws* rather than returning `null`, so that guard never ran.

### AI assistance

Models: Claude Opus 5

### Validation

- [ ] `bun run check`
- [x] Tests added or updated, or not needed because: explain here
- [x] CHANGELOG.md updated, or not needed because: explain here

**On `bun run check`** — it does not pass end-to-end in my environment, for reasons unrelated to this change:

- `check:secrets` fails because neither `gitleaks` nor `go` is installed locally. This change introduces no credentials.
- 78 engine tests fail on this machine **and fail identically on unmodified `master`**: the `.fig` fixtures under `tests/fixtures/` are unfetched git-LFS pointer files, so `unzipSync` rejects them. None of the failures reference the code touched here.

Everything else passes: `lint` (0 errors), `tsgo --noEmit`, `check:vue`, `check:arch`, `check:monorepo`, `check:i18n`, `check:packages`, `test:type-shapes`, `test:dupes`.

**Tests** — three regression tests added to `tests/engine/render/canvas/retained-backing.test.ts`:

- the backing stays within `MAX_TEXTURE_SIZE` at the reported 2998×1490 @ dpr 2 geometry;
- the margin is *not* shrunk when it already fits (asserts the exact `11514×8940` request, so the clamp cannot silently over-trim);
- `renderSceneBacking` does not propagate a throwing `makeSurface`.

I confirmed these fail without the fix — reverting the clamp makes the first test fail with `Received: 17988`, matching the value measured in the browser.

**Manual verification** — reproduced and confirmed fixed in the running app: drew a rectangle, toggled the UI three times, and checked the console after each. Zero errors, rendering intact, panels/layers/selection preserved, and tab switching still behaves correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Keep the canvas rendering when hiding or showing the UI: the editor no longer remounts the canvas between layouts, and the retained scene backing stays within the GPU texture-size limit on wide HiDPI viewports.
 - Match regional browser languages to supported locales without selecting a secondary language. (#417)
 - Save auto-layout frames that stretch their children to `.fig` without failing. (#427)
 - Reduce large `.fig` page-switch work to the active page, reuse fixed-point propagation indexes, and coalesce Layers tree rebuilds. (#420)
--- a/packages/core/src/canvas/renderer.ts
+++ b/packages/core/src/canvas/renderer.ts
@@ -62,6 +62,19 @@ export interface PendingFontNode {
 
 import type { RenderOverlays, RulerTheme } from './renderer/types'
 
+/** Conservative floor mandated by the WebGL2 spec when the real limit is unknown. */
+const DEFAULT_MAX_TEXTURE_SIZE = 2048
+
+function readMaxTextureSize(gl: WebGL2RenderingContext | null): number {
+  if (!gl) return DEFAULT_MAX_TEXTURE_SIZE
+  try {
+    const size: unknown = gl.getParameter(gl.MAX_TEXTURE_SIZE)
+    return typeof size === 'number' && size > 0 ? size : DEFAULT_MAX_TEXTURE_SIZE
+  } catch {
+    return DEFAULT_MAX_TEXTURE_SIZE
+  }
+}
+
 export class SkiaRenderer {
   ck: CanvasKit
   surface: Surface
@@ -173,6 +186,8 @@ export class SkiaRenderer {
   panY = 0
   zoom = 1
   dpr = 1
+  /** GPU limit for a single texture dimension, in device pixels. */
+  maxTextureSize = DEFAULT_MAX_TEXTURE_SIZE
   viewportWidth = 0
   viewportHeight = 0
   showRulers = true
@@ -412,6 +427,7 @@ export class SkiaRenderer {
   constructor(ck: CanvasKit, surface: Surface, gl?: WebGL2RenderingContext | null) {
     this.ck = ck
     this.surface = surface
+    this.maxTextureSize = readMaxTextureSize(gl ?? null)
     this.profiler = new RenderProfiler(ck, gl ?? null)
     initializeRendererPaints(this)
   }
--- a/packages/core/src/canvas/renderer/retained-backing.ts
+++ b/packages/core/src/canvas/renderer/retained-backing.ts
@@ -153,11 +153,24 @@ function drawSceneBacking(
   return true
 }
 
+function maxSceneBackingPixels(r: SkiaRenderer): number {
+  return Math.max(1, r.maxTextureSize)
+}
+
 function sceneBackingGeometry(r: SkiaRenderer) {
+  // The backing texture is SCENE_BACKING_SCALE times the viewport in each axis,
+  // then multiplied again by dpr in createSceneBackingSurface. On a wide
+  // viewport with a HiDPI display that product overshoots the GPU's maximum
+  // texture dimension (commonly 16384), which makes the allocation fail. Cap
+  // the margin so the final pixel size always stays inside the limit — a
+  // smaller backing only costs extra re-records while panning.
+  const maxBackingPx = maxSceneBackingPixels(r)
+  const maxWidth = Math.max(1, Math.floor(maxBackingPx / r.dpr))
+  const maxHeight = Math.max(1, Math.floor(maxBackingPx / r.dpr))
   const marginX = r.viewportWidth * ((SCENE_BACKING_SCALE - 1) / 2)
   const marginY = r.viewportHeight * ((SCENE_BACKING_SCALE - 1) / 2)
-  const width = Math.max(1, Math.ceil(r.viewportWidth + marginX * 2))
-  const height = Math.max(1, Math.ceil(r.viewportHeight + marginY * 2))
+  const width = clamp(Math.ceil(r.viewportWidth + marginX * 2), 1, maxWidth)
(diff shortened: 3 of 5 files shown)


dannote:
Thanks for the detailed investigation—the retained-backing failure is real, but I don't think the Teleport/`v-show` layout rewrite is the right fix for it.

The crash is caused by an optional retained cache requesting a 17988×8940 offscreen surface for the reported viewport and then letting CanvasKit's allocation exception escape. `EditorCanvas` remounting across genuinely different desktop/mobile/collapsed/no-chrome layouts is normal lifecycle behavior; keeping both layout trees mounted and moving a single canvas through ref-driven Teleports adds substantial shell complexity and couples layout ownership to renderer lifetime.

I've opened #469 as a renderer-only replacement. It bounds retained backing memory, preserves the full margin where affordable, and falls back to direct scene rendering when CanvasKit rejects the cache surface. It includes the reported 2998×1490 @ DPR 2 regression and a throwing-allocation test.

dannote:
Closing in favor of the smaller renderer-level replacement in #469.

Issue #436, opened 2026-07-31
Anthropic provider unusable in web build: missing `anthropic-dangerous-direct-browser-access` header


rcoenen:
### Description

The `anthropic` provider cannot work in the web build, because OpenPencil never sends the header Anthropic requires to enable browser access. Anthropic **does** support direct browser calls — it just gates them behind an explicit opt-in header that neither `@ai-sdk/anthropic` nor our adapter sets.

Users hit `connectionTestBrowserNetworkFailed` ("Could not reach this endpoint from the browser…") and reasonably conclude the API is server-only.

### Evidence

Anthropic returns no CORS header by default, but returns one as soon as the opt-in header is present:

```console
$ curl -sD - -o /dev/null -X POST https://api.anthropic.com/v1/messages \
    -H "Origin: https://openpencil.dev" -H "content-type: application/json" -d '{}'
HTTP/2 401
(no access-control-allow-origin)

$ curl -sD - -o /dev/null -X POST https://api.anthropic.com/v1/messages \
    -H "Origin: https://openpencil.dev" -H "content-type: application/json" \
    -H "anthropic-dangerous-direct-browser-access: true" -d '{}'
HTTP/2 401
access-control-allow-origin: *
access-control-expose-headers: *
```

`@ai-sdk/anthropic@3.0.74` sends only `x-api-key` and `anthropic-version` — the string `dangerous-direct-browser-access` does not appear anywhere in the package. `createAnthropicCompatibleAdapter` in `src/app/ai/providers/compatible.ts` doesn't add it either.

`zai` masks the problem: it uses the same adapter but its gateway sends `Access-Control-Allow-Origin` unconditionally, so it works while direct `anthropic` does not.

### Proposed fix

```ts
// src/app/ai/providers/compatible.ts, createAnthropicCompatibleAdapter
const provider = createAnthropic({
  apiKey: config.apiKey,
  baseURL: resolveEndpoint(options.baseURL, config),
  fetch: runtime.fetch,
  headers: IS_TAURI ? undefined : { 'anthropic-dangerous-direct-browser-access': 'true' }
})
```

Gating on `IS_TAURI` avoids advertising browser access from the desktop build, which routes through `tauriFetch` and isn't subject to CORS.

Note the header name is Anthropic's own wording: it signals that the user's API key will live in browser storage. That's inherent to BYOK in a web app and already the case for every other provider we support, but it's worth a deliberate decision rather than an implicit one.

### Related

[...]

PR #437, opened 2026-07-31
docs: add BYOK provider & model compatibility reference

rcoenen:
### Summary

OpenPencil has no backend, so BYOK requests go from the browser straight to the provider. That makes two things matter that aren't documented anywhere — whether a provider sets CORS headers on the **response** (not just the preflight), and whether a model's **streaming tool calls** are well-formed enough for the AI SDK to parse.

Both vary widely between providers, and when either fails the user sees a generic error that looks like a wrong API key. This adds a reference page recording what has actually been measured, structured so others can add rows.

Related to #436, which covers one specific instance of this (the Anthropic browser header).

### What changed

- Add `docs/byok-providers.md`:
  - **Provider CORS table** — 11 providers, each dated. Most work in the browser today; Anthropic needs a header we don't send yet (#436); TensorX sets CORS on the preflight but not the response.
  - **Model tool-calling table** — 5 models scored on whether they emit calls, whether streamed `id`s are correct, and whether concatenated arguments parse as JSON.
  - **Two documented failure modes** — vLLM misrouting a tool call's final argument chunk to `index + 1` without an `id` (crashes the stream *and* leaves arguments unterminated), and reasoning models returning `null` content under the connection test's `maxOutputTokens: 1`.
  - **Copy-paste curl recipes** so contributors can produce comparable rows, including the "check the POST, not the OPTIONS" trap that makes CORS look fine when it isn't.
- Link the page from the README's *Built-in chat* section.

Every row carries a `Tested` date, and the page states up front that stale rows should be treated as unverified — provider behaviour changes without announcement.

### AI assistance

Models: Claude Opus 5

### Validation

- [ ] `bun run check` — not run. Docs-only change (two `.md` files, no code, no types, no i18n keys). Ran `oxfmt` on both files, which is what CI enforces for formatting.
- [x] Tests added or updated, or not needed because: documentation-only; there is no behaviour to test. The findings themselves were produced by live requests against each provider, and the page documents how to reproduce them.
- [ ] CHANGELOG.md updated, or not needed because: no user-facing behaviour change — this documents existing behaviour rather than altering it.

Note: `bun run check:secrets` cannot run in my environment (neither `gitleaks` nor `go` is installed, so it fails on clean `master` too). I verified manually that the commit contains no credentials — the only key-shaped strings in the doc are the placeholders `KEY="sk-…"` and `$KEY`.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on BYOK provider and model compatibility.
  * Documented browser CORS support, streaming tool-call quality, pricing, models.dev limitations, vLLM and reasoning-model considerations, testing procedures, regional catalog differences, contribution guidance, and security practices.
  * Updated the README with a link to the new compatibility guidance.
  * Reformatted the Quality gates and Tech stack tables for improved readability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,77 +2,86 @@
 
 ## Unreleased
 
-### Fixed
+### Breaking changes
 
-- Match Figma Plugin API vector path and network editing, including bounds, winding rules, region fills, validation, and handle mirroring. (#444)
-- Let AI and MCP tools create arbitrary vectors from SVG path data without leaving blank layers after invalid input. (#440)
-- Show stroke colors and weights in AI visual descriptions. (#447)
-- Stop warning AI agents that supported inline SVG attributes were ignored. (#445)
-- Help AI agents discover every shape supported by `create_shape`. (#448)
-- Keep `fill="none"` and `stroke="none"` SVG paths transparent when rendering inline artwork. (#446)
-- Match regional browser languages to supported locales without selecting a secondary language. (#417)
-- Save auto-layout frames that stretch their children to `.fig` without failing. (#427)
-- Preserve multiple colors in imported vector artwork such as multi-color logos. (#386)
-- Edit vectors in opened documents at the correct position, with live fills and undo/redo. (#390)
-- Reduce large `.fig` page-switch work to the active page, reuse fixed-point propagation indexes, and coalesce Layers tree rebuilds. (#420)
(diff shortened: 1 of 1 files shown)


dannote:
Superseded by #464. The replacement publishes the reference in the canonical VitePress docs, links all maintained AI Chat locales to it, updates stale product behavior, and addresses the open reproducibility/security review findings.

PR #443, opened 2026-07-31
feat(app): add a Back to workspace button to the editor header

rcoenen:
### Problem

Once a document is opened from the storage workspace, there is no way back to it. The only in-app route is Settings → Cloud storage → Open workspace, and the desktop shell has no browser back at all.

Browser back is also not a substitute even on web: it is history-dependent, so it may or may not land on the workspace depending on how you got there. "Back to workspace" is a fixed destination.

### Change

Adds an explicit button at the top of the editor header that routes to `/storage`.

It lives in the header row that renders on **both** web and Tauri — the `File/Edit/View` menubar below it is `v-if="!IS_TAURI"`, since desktop uses the native menu — so this is one button with identical behaviour on both platforms.

Shown unconditionally rather than gated on storage being configured: `StorageView` already prompts for setup when it is not, and hiding navigation based on config is how the feature stays undiscovered.

### Screenshot

Button appears above the document name in the left panel; clicking it navigates to the workspace grid.

---

Model disclosure: implemented with Claude Opus 5.

## Summary by CodeRabbit

* **New Features**
  * Added a “Back to workspace” button above the document header.
  * The button returns users to the storage workspace and includes localized text and a back-arrow icon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,77 +2,86 @@
 
 ## Unreleased
 
-### Fixed
+### Breaking changes
 
-- Match Figma Plugin API vector path and network editing, including bounds, winding rules, region fills, validation, and handle mirroring. (#444)
-- Let AI and MCP tools create arbitrary vectors from SVG path data without leaving blank layers after invalid input. (#440)
-- Show stroke colors and weights in AI visual descriptions. (#447)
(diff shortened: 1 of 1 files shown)


dannote:
Superseded by #463 after visual review. The replacement removes the full-width editor-header control and exposes **File → Open storage workspace…** through the shared browser/native menu architecture.

Issue #442, opened 2026-07-31
Plugin API gaps in the eval sandbox: rescale(), page.backgrounds, absoluteRenderBounds


rcoenen:
Three smaller gaps in the same `eval` / Figma-plugin-API surface as #441, grouped since they share a cause. All observed in agent debug logs on `master` @ `7e8feb99`.

### 1. `rescale()` is not implemented

```js
star.rescale(760 / 520)   // → TypeError: star.rescale is not a function
group.rescale(3.3)        // → TypeError: group.rescale is not a function
```

It is part of the Figma plugin API and is the natural way to scale a node or group proportionally. Agents reach for it, fail, then fall back to `node_resize` — which resizes the container without scaling children, so a scaled group ends up with unscaled contents.

### 2. `page.backgrounds` returns nothing

```js
return page.backgrounds                                  // → no return value
return JSON.stringify(page.backgrounds)                  // → no return value
return JSON.parse(JSON.stringify({ bg: page.backgrounds })) // → {}
```

An agent trying to compose against the canvas colour (for example deciding whether a masking rectangle can pass as background) has no way to read it.

### 3. `absoluteRenderBounds` returns the node box, not rendered bounds

For vector nodes with no geometry it reported `100x100` — the node's nominal size — rather than the actual rendered extent, which was empty. Since the documented purpose of `absoluteRenderBounds` is the true rendered extent including strokes and effects, this makes it unusable for verifying that something drew.

This one compounds #441: with geometry silently dropped **and** bounds reporting a plausible-looking box, an agent has no signal at all that its drawing failed.

### Suggested handling

Implement `rescale()` and `page.backgrounds`; make `absoluteRenderBounds` reflect real geometry, or drop it rather than return a misleading value. As with #441, anything intentionally unsupported should throw rather than return a plausible wrong answer.

dannote:
Thanks for the detailed report, Rob. Some related foundations have since landed:

- #317 made interactive resizing scale the children of groups and boolean operations.
- #278 preserved canvas background colors and paints when importing and exporting `.fig` files.

Those changes do not yet complete the requested Plugin API surface:

- `node.rescale()` is still unavailable in `eval`; #317 only covers editor resize interactions.
- `figma.currentPage.backgrounds` is not exposed, despite the background metadata now being preserved internally.
- `absoluteRenderBounds` still returns the nominal node rectangle, so an empty vector can report plausible non-empty bounds.

[...]

Issue #439, opened 2026-07-31
create_vector rejects SVG path data, so AI agents cannot draw custom shapes


rcoenen:
### Summary

`create_vector` is the only tool that accepts arbitrary geometry, but its `path` parameter requires OpenPencil's internal `VectorNetwork` JSON — an indexed graph structure that no LLM has in its training data, and which the tool schema doesn't describe. In practice this means **AI agents cannot draw custom shapes at all**.

### What happens

Asking the design agent for any non-primitive shape ("a 3D embossed star behind this character") ends with nothing drawn. The model reaches for `create_vector`, its `path` payload fails validation, and the node is created with no geometry.

From a real session, the model diagnosing it in its own words:

> `createVector` won't accept path data this way (paths count = 0) — that's why nothing draws.

It then fell back to rasterising the shape by hand — computing per-scanline intersections with the star polygon and emitting ~52 `Line` nodes per layer, ~120 nodes total. The result rendered as a rectangular barcode, not a star.

Tool output showing the resulting nodes:

```json
{ "name": "StarLinesBase",   "type": "VECTOR", "w": 100, "h": 100, "strokeW": 4, "paths": 0 }
{ "name": "StarLinesYellow", "type": "VECTOR", "w": 100, "h": 100, "strokeW": 4, "paths": 0 }
```

### Why it fails

`packages/core/src/tools/create/vector.ts`:

```js
path: { type: 'string', description: 'VectorNetwork JSON' }
```

That is the entire specification given to the model. The actual required structure is an indexed graph:

```
vertices: [{x, y}, …]
segments: [{start: <vertexIndex>, end: <vertexIndex>,
            tangentStart: {x,y}, tangentEnd: {x,y}}, …]
regions:  [{windingRule, loops: [[<segmentIndex>, …]]}]
```

Segments reference vertices by array index, regions reference segments by array index. A model can emit this correctly — but not by guessing it from the words "VectorNetwork JSON". Every model defaults to an SVG path `d` string, which fails `safeDestr` or `validateVectorNetwork`.

### Scope

This affects both tool surfaces, since `packages/mcp/src/tool/registration.ts` imports `ALL_TOOLS` from `@open-pencil/core/tools`:

- the in-app AI chat panel
- any external MCP client driving OpenPencil (Claude Code, etc.)

### Suggested fix

Accept an SVG path `d` string in addition to `VectorNetwork` JSON. The parser already exists and returns exactly the right type:

[...]

PR #444, opened 2026-07-31
fix(figma-api): allow vectorPaths and vectorNetwork to be assigned

### LINKED ISSUE #441: eval sandbox silently discards vectorPaths assignment, so agents cannot draw custom geometry
### Summary

In the `eval` sandbox, `node.vectorPaths = [...]` is silently discarded, and `vectorNetwork` is not exposed on the node proxy at all. Agents writing ordinary Figma-plugin-style code get no error, no warning, and a vector node with zero geometry that renders nothing.

### Cause

`packages/core/src/figma-api/accessors/vector.ts:13` defines `vectorPaths` with a getter and no setter:

```js
Object.defineProperty(prototype, 'vectorPaths', {
  get(this: ProxyThis): readonly FigmaVectorPath[] { … }
})
```

In non-strict code, assigning to a getter-only property is a silent no-op. `packages/core/src/figma-api/proxy.ts:108` likewise declares it `readonly`. `vectorNetwork` has no accessor, so assigning to it just attaches an inert own-property.

### Observed behaviour

From an agent debug log, asked to draw a sailboat outline. It created six vectors and set `vectorPaths` on each:

```js
const v = figma.createVector()
v.vectorPaths = [{ windingRule: "NONE", data: "M 50 6 L 50 70 Q 45 72 18 74 L 50 6 Z" }]
```

Result — every node reports 100×100 with no path data:

```
[{"name":"MainSail","w":100,"h":100,"hasPath":false},
 {"name":"JibSail","w":100,"h":100,"hasPath":false},
 {"name":"Mast","w":100,"h":100,"hasPath":false},
 {"name":"Flag","w":100,"h":100,"hasPath":false},
 {"name":"Hull","w":100,"h":100,"hasPath":false},
 {"name":"Waves","w":100,"h":100,"hasPath":false}]
```

Reading it back gives `data: "none"`. Setting `vectorNetwork` directly instead returns `paths: 0`.

The agent's own conclusion, after four attempts across two rebuilds:

> Path data is being dropped silently.

> The sandbox can't set custom path data.

It then abandoned real geometry and approximated the drawing with stacked `Line`/`Polygon`/`Ellipse` primitives plus an opaque white rectangle as a fake mask.

### Why this matters

Silent failure is the expensive part. A hard error on the first assignment would cost one step; instead the agent re-derived, rebuilt, and re-verified across roughly a dozen tool calls before concluding the capability was missing — and in an earlier session produced ~120 `Line` nodes hand-rasterising a shape for the same reason.

### Suggested fix

Either:

1. Implement the setter — parse the `d` string via `parseSVGPath` (`packages/scene-graph/src/parse-path.ts:17`, which returns `VectorNetwork` directly) and assign to the node, and expose a `vectorNetwork` accessor with a setter; or
2. If these are intentionally read-only, **throw** on assignment rather than ignoring it, so the caller learns immediately.

(1) is preferable — it is the documented Figma plugin API surface, and agents assume it works.

### Related

- #439 / #440 — the `create_vector` tool had the mirror-image problem on the tool surface.

### Environment

`master` @ `7e8feb99`.

rcoenen:
Fixes #441

### Problem

`vectorPaths` was defined with a getter and no setter (`packages/core/src/figma-api/accessors/vector.ts:13`), so in the `eval` sandbox `node.vectorPaths = [...]` was silently discarded. `vectorNetwork` had no accessor at all, so assigning to it attached an inert own-property.

An agent writing ordinary Figma-plugin-style code got a named node with zero geometry, no error, and nothing rendered.

### Change

Setters for both, following the existing `fills`/`strokes` pattern in `accessors/visual.ts`:

- **`vectorPaths`** — parses each `d` string through `parseSVGPath` and merges them into a single network with offset vertex/segment indices, so a multi-path assignment lands as one node's geometry.
- **`vectorNetwork`** — validates first and throws on out-of-range indices rather than storing something that cannot render. `null` clears it.

Both clear `fillGeometry`, whose command blobs are derived from the previous network and would otherwise keep painting the old shape.

The proxy declarations for both are no longer `readonly`.

### Why the silence mattered

A hard error on the first assignment would have cost one step. Instead agents re-derived, rebuilt, and re-verified across roughly a dozen tool calls before concluding the capability was missing — and in one session emitted ~120 `Line` nodes hand-rasterising a shape by scanline. The full transcript is in #441.

### Tests

Three added in `tests/engine/figma/api/vector/paths.test.ts`:

- assigning `vectorPaths` produces real geometry, readable back out
- assigning multiple paths merges into one network with both regions intact
- assigning `vectorNetwork` sets geometry, and out-of-range indices throw

All three fail on `master` and pass with this change (verified by reverting the fix and re-running).

### Related

- #439 / #440 — the mirror-image problem on the `create_vector` tool surface
- #442 — remaining gaps in the same plugin-API surface

---

Model disclosure: implemented with Claude Opus 5.

## Summary by CodeRabbit

* **New Features**
  * Added editing support for vector paths and networks, including regions, fills, open paths, winding rules, and handle mirroring.
  * Added configurable precision for SVG export and improved conversion of connected segments.

* **Bug Fixes**
  * Improved vector geometry normalization, bounds handling, validation, and Figma-compatible readback.
  * Improved parsing and error reporting for invalid vector data.
  * Vector nodes now default to no handle mirroring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Match Figma Plugin API vector path and network editing, including bounds, winding rules, region fills, validation, and handle mirroring. (#444)
 - Let AI and MCP tools create arbitrary vectors from SVG path data without leaving blank layers after invalid input. (#440)
 - Show stroke colors and weights in AI visual descriptions. (#447)
 - Stop warning AI agents that supported inline SVG attributes were ignored. (#445)
--- a/packages/core/src/figma-api/accessors/vector.ts
+++ b/packages/core/src/figma-api/accessors/vector.ts
@@ -1,29 +1,317 @@
-import { raw, type NodeProxyInternals, type ProxyThis } from '#core/figma-api/accessor-utils'
+import {
+  mergeVectorNetworks,
+  normalizeVectorNetwork,
+  transformVectorNetwork,
+  validateVectorNetwork
+} from '@open-pencil/scene-graph'
+import type {
+  Fill,
+  GeometryPath,
+  HandleMirroring,
+  VectorNetwork as SceneVectorNetwork,
+  VectorSegment,
+  VectorVertex,
+  WindingRule
+} from '@open-pencil/scene-graph'
+import { copyFills } from '@open-pencil/scene-graph/copy'
+import { parsePluginVectorPath } from '@open-pencil/scene-graph/parse-path'
+
+import {
+  raw,
+  updateNode,
+  type NodeProxyInternals,
+  type ProxyThis
+} from '#core/figma-api/accessor-utils'
 import { geometryBlobToSVGPath, vectorNetworkToSVGPaths } from '#core/io/formats/svg/paths'
+import { computeAccurateBounds } from '#core/vector/curve-math'
+import { regenerateFillGeometry } from '#core/vector/fill-geometry'
 
 export interface FigmaVectorPath {
-  readonly windingRule: 'NONZERO' | 'EVENODD'
+  readonly windingRule: WindingRule | 'NONE'
   readonly data: string
 }
 
-export function installVectorNodeProxyAccessors(
-  prototype: object,
-  internals: NodeProxyInternals
+interface FigmaVectorRegion {
+  readonly windingRule: WindingRule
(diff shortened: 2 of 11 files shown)


dannote:
@coderabbitai full review

PR #446, opened 2026-07-31
fix(icons): honour fill="none" and stroke="none"

rcoenen:
## Problem

`createIconFromPaths` gates painting on truthiness:

```ts
if (path.fill) {
  const fillColor = path.fill === 'currentColor' ? color : parseColor(path.fill)
```

`"none"` is a non-empty string, so it takes the paint branch. `parseColor("none")` does not parse and falls back to black, so `fill="none"` produces an opaque black fill.

Inline `<svg>` in design JSX runs through this same pipeline — `renderSvgNode` builds an `IconData` and calls `createIconFromPaths` — so every outline drawing comes back as solid black shapes sitting on top of their own strokes:

```jsonc
// node rendered from <path stroke="#021A3B" stroke-width="6" fill="none" />
{ "strokes": [{ "color": navy, "weight": 5.14 }],
  "fills":   [{ "type": "SOLID", "color": { "r": 0, "g": 0, "b": 0, "a": 1 } }] }
```

The vectorize path already gets this right — `svg/to-vectors.ts` guards with `path.fill !== 'none'` in both `resolveFill` and `resolveStrokes`. This just brings the icon path in line.

## Why it matters

`describe` then summarises the node by its fill, reporting `#000000`. An agent drawing a sailboat outline read that back, concluded its stroke props had not applied, patched fills by hand through `eval`, second-guessed the result and deleted it.

## Fix

Skip the paint when the value is `"none"`, for both fill and stroke.

## Test

`tests/engine/icons/render.test.ts` — new file, three cases. Fails on master (2 of 3), passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved icon rendering by correctly ignoring `none` fill and stroke values.
  * Preserved valid fill and stroke colors for accurate icon appearance.

* **Tests**
  * Added coverage for icons with absent and valid fill or stroke colors.

### DIFF
--- a/packages/core/src/icons/render.ts
+++ b/packages/core/src/icons/render.ts
@@ -33,7 +33,10 @@ export function createIconFromPaths(
     vector.x = 0
     vector.y = 0
 
-    if (path.fill) {
+    // `fill="none"` is a non-empty string, so a truthiness check treats it as a
+    // colour and parseColor() falls back to black — outline drawings came out
+    // as solid black blobs.
+    if (path.fill && path.fill !== 'none') {
       const fillColor = path.fill === 'currentColor' ? color : parseColor(path.fill)
       graph.updateNode(vector.id, {
         fills: [{ type: 'SOLID', color: fillColor, opacity: 1, visible: true }]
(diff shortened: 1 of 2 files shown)


dannote:
Thanks, Rob! This has been adapted and landed on `master` as `f047b0bc`, with your co-author credit preserved. I’m closing the superseded PR.

PR #455, opened 2026-07-31
feat(chat): switch models from the chat panel

rcoenen:
### Summary

The model chip above the chat input is a dropdown for built-in providers, but a **static label** for OpenAI-compatible and custom-model setups:

```vue
<template v-else-if="isCustomProvider || usesCustomModel">
  <div ...>{{ selectedModelName }}</div>   <!-- not a control -->
</template>
<ProviderModelSelect v-else>               <!-- built-ins get a dropdown -->
```

So with several models already configured, anyone on a custom endpoint has to open Settings → AI & agents → Assignments → Design agent every time they want to switch. That's four clicks and a modal to change one field that's already visible in the chat.

This adds a switcher in the chat panel that reassigns the design role directly.

### What changed

- **`ChatProfileSelect.vue`** (new) — dropdown over saved model profiles, filtered to those with the `tools` capability, calling `setModelRoleAssignment('design', profileId)`. Vision-capable profiles get a badge so it's clear which ones accept images.
- **`ChatInput.vue`** — renders the switcher ahead of the static-label branch.
- **Two tests** in `tests/engine/app/ai/model-roles.test.ts` covering the invariants the dropdown relies on: switching the design role between profiles resolves the new profile *and* its connection, and a profile without `tools` is refused.

### Design notes

- **Only renders with 2+ tools-capable profiles.** With one configured model a dropdown is noise, so the existing label is kept.
- **Shows the profile name** ("Nebius Kimi3") rather than the raw model ID ("moonshotai/Kimi-K3"), matching what the user typed in Settings.
- **Writes the same `assignments.design`** field the Settings dropdown writes, so the two views stay in sync with no extra state.
- **Built-in providers are untouched** — `ProviderModelSelect` still handles those. Their chip switches `modelID` within one connection, which is different behaviour from switching profiles, and I didn't want to change it in the same PR.
- Reuses the existing `useSelectUI` styling and the `dialogs.models` / `dialogs.modelCapabilityVisionShort` strings, so no new i18n keys.

### AI assistance

Models: Claude Opus 5

### Validation

- [x] `bunx tsgo --noEmit` clean, `bun run check:vue` clean, `oxlint` 0 warnings / 0 errors, `oxfmt` applied
- [x] Tests added: `tests/engine/app/ai/model-roles.test.ts` — 10 pass / 0 fail
- [x] Manually verified in the dev server with two configured profiles (Nebius Kimi-K3 and Scaleway gemma-4): the dropdown appears, switching updates the chat agent, and the Settings assignment follows
- [ ] CHANGELOG.md updated, or not needed because: happy to add an entry if you'd like one for a UI addition of this size

Full `bun run check` not run locally — `check:secrets` needs `gitleaks` or `go`, neither installed here (it fails on clean `master` too). Ran the type, lint, format and test stages this change touches.

## Summary by CodeRabbit

* **New Features**
  * Added a profile selector for choosing among compatible AI models during chat.
  * Displays the selected design profile and highlights vision-capable profiles.
  * Preserves existing model selection behavior when profile switching is unavailable.
* **Bug Fixes**
  * Improved fallback handling when a selected profile does not support required tools.
* **Tests**
  * Added coverage for switching profiles and resolving their connections.

### DIFF
--- a/src/components/chat/ChatInput.vue
+++ b/src/components/chat/ChatInput.vue
@@ -2,11 +2,13 @@
 import { TooltipProvider } from 'reka-ui'
 import { computed, ref } from 'vue'
 
+import ChatProfileSelect from '@/components/chat/ChatProfileSelect.vue'
 import ProviderModelSelect from '@/components/chat/ProviderModelSelect.vue'
 import AppInput from '@/components/ui/AppInput.vue'
 import Tip from '@/components/ui/Tip.vue'
 import { useButtonUI } from '@/components/ui/button'
 import { useAIChat } from '@/app/ai/chat/use'
+import { aiModelSettings, designModelProfile } from '@/app/ai/models'
 import { openSettingsDialog } from '@/app/settings/dialog'
 import { useI18n } from '@open-pencil/vue'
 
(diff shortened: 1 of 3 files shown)


dannote:
Superseded by #468, which rebases the profile switcher onto current master, centralizes Design-profile eligibility, preserves built-in provider model selection, adds localized accessibility, and includes local visual review.

PR #465, opened 2026-08-03
fix(fig): scale target-aspect instance subtrees

zackarychapple:
## Summary

- Use preserved Figma `targetAspectRatio` metadata to continue scaling through fixed wrapper layers.
- Scale only branches that lead to descendants with explicit `SCALE` constraints.
- Preserve fixed, pinned, centered, and unrelated siblings inside the same subtree.
- Add regression coverage modeled on a 310×62 vector logo instantiated at 100×20 and ordinary resized icons.
- Document the import-fidelity fix in the Unreleased changelog.

## Root cause

Constraint scaling stopped at a fixed wrapper frame, so scale-constrained vector paths nested below it retained the component's original geometry and rendered oversized or clipped.

The initial fix used `targetAspectRatio` as permission to scale the entire populated subtree. Real-file validation showed that was too broad: Preline contains 1,297 target-aspect instances, and whole-subtree scaling changed 537 vector nodes, including icons that were already correct.

The adapted implementation uses target-aspect metadata only to allow traversal through fixed wrappers. Each branch is cached and traversed only when it contains an explicit horizontal or vertical `SCALE` descendant. The component remains the geometry basis, preventing ordinary resized icons from being scaled twice.

## Impact

Proportion-constrained `.fig` instances such as logos retain their intended size when vector geometry is nested under fixed wrapper layers, without changing unrelated target-aspect icons or fixed sibling geometry.

## Validation

- `bun run check`
- `bun --cwd packages/fig test` — 33 passed
- Shadcn fixture — no affected geometry
- Preline fixture — 0 geometry changes across 427,980 nodes
- Duplication detection — 0 clones

## Attribution

Zack Chapple authored the original target-aspect scaling contribution. The maintainer adaptation narrows its traversal semantics against current `master` and adds real-file regression coverage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved the intended sizing and proportions of imported logos and icons.
  - Improved scaling for proportion-constrained graphics, including nested elements, vector content, and fixed-size wrappers.
  - Preserved inset and constraint geometry when resizing graphics to a target aspect ratio.
  - Applied scaling consistently to descendants with scale constraints.
  - Ignored invalid or missing sizing metadata to prevent incorrect geometry changes.

- **Documentation**
  - Added an Unreleased changelog entry describing the sizing improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,9 @@
 
 ### Fixed
 
-- Match Figma auto-layout spacing, padding, min/max constraints, scalar variable bindings, CanvasKit-shaped generated text, imported text bounds, and nested instance geometry more closely.
-- Match Figma Plugin API vector path and network editing, including bounds, transforms, winding rules, region fills, validation, and handle mirroring. (#444)
+- Scale proportion-constrained `.fig` instance geometry through fixed wrapper layers so imported logos and icons retain their intended size.
+- Match Figma auto-layout spacing, padding, min/max constraints, scalar variable bindings, imported text bounds, and nested instance geometry more closely.
+- Match Figma Plugin API vector path and network editing, including bounds, winding rules, region fills, validation, and handle mirroring. (#444)
 - Let AI and MCP tools create arbitrary vectors from SVG path data, validating input without leaving blank layers behind. (#440)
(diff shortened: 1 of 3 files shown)


dannote:
Thank you!

PR #469, opened 2026-08-05
fix(canvas): bound retained backing allocations

### LINKED ISSUE #432: Canvas goes blank with "Cannot set properties of null (setting 'be')" when toggling the UI

dannote:
Fixes #432.
Supersedes #433 with a renderer-only fix.

## Summary

The reported failure is real: hiding the UI widens the viewport, and the retained scene cache can request an enormous offscreen CanvasKit surface. At 2998×1490 CSS px and DPR 2, the fixed 3× backing requests 17988×8940 device pixels. CanvasKit may throw while allocating it, blanking the scene layer.

This PR fixes that allocation boundary without restructuring `EditorView`:

- Scale the retained margin down to a 16M-device-pixel budget on large viewports.
- Keep the current 3× retained margin when it fits the budget.
- Catch offscreen `makeSurface` allocation failures and let the existing render pipeline draw the scene directly for that frame.
- Add regression coverage for the reported geometry and throwing CanvasKit behavior.

## Why not the layout rewrite from #433?

`EditorCanvas` remounting is normal lifecycle behavior across the editor's genuinely different desktop, mobile, collapsed, and no-chrome layouts. Moving one canvas through ref-driven Teleports while keeping the desktop splitter hidden with `v-show` adds two permanent layout trees and tightly couples canvas ownership to view layout. It also broadens a renderer allocation bug into a risky shell rewrite.

The crash occurs because an optional cache allocation is unbounded and throws. Fixing that boundary preserves the simpler layout and existing canvas lifecycle.

## Validation

- `bun test tests/engine/render/canvas/retained-backing.test.ts` — 7 passing
- `bun run check:arch`
- `bun run check` — passes; existing unrelated lint warnings remain in `props-overrides.ts` and `effective-generated-text.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved canvas rendering when showing or hiding the UI on wide, high-resolution displays.
  * Prevented excessive graphics memory allocation during complex scene rendering.
  * Added a graceful fallback for devices that cannot create required offscreen surfaces.
  * Prevented repeated allocation attempts after a failure, improving rendering stability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 
 ### Fixed
 
+- Keep the canvas rendering when hiding or showing the UI on wide HiDPI viewports by bounding retained scene backing allocations and falling back to direct rendering when CanvasKit rejects an offscreen surface.
 - Match Figma auto-layout spacing, padding, min/max constraints, scalar variable bindings, CanvasKit-shaped generated text, imported text bounds, and nested instance geometry more closely.
 - Match Figma Plugin API vector path and network editing, including bounds, transforms, winding rules, region fills, validation, and handle mirroring. (#444)
 - Let AI and MCP tools create arbitrary vectors from SVG path data, validating input without leaving blank layers behind. (#440)
--- a/packages/core/src/canvas/renderer.ts
+++ b/packages/core/src/canvas/renderer.ts
@@ -122,6 +122,7 @@ export class SkiaRenderer {
   } | null = null
   sceneBackingPreviewUntil = 0
   sceneBackingNeedsCrispRender = false
+  sceneBackingAllocationFailed = false
   sceneBackingBuild: {
     surface: Surface
     graph: SceneGraph
@@ -451,6 +452,7 @@ export class SkiaRenderer {
   replaceSurface(surface: Surface): void {
     this.surface.delete()
     this.surface = surface
+    this.sceneBackingAllocationFailed = false
     this.invalidateScenePicture()
   }
 
--- a/packages/core/src/canvas/renderer/retained-backing.ts
+++ b/packages/core/src/canvas/renderer/retained-backing.ts
@@ -10,6 +10,7 @@ import type { RenderLayer } from './pipeline'
 
 const now = typeof performance !== 'undefined' ? () => performance.now() : () => 0
 const SCENE_BACKING_SCALE = 3
+const MAX_SCENE_BACKING_DEVICE_PIXELS = 16_000_000
 const FRAME_BUDGET_60HZ_MS = 1000 / 60
 const MIN_SCENE_BACKING_IDLE_FRAMES = 2
 const MAX_SCENE_BACKING_IDLE_FRAMES = 18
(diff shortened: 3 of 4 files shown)


dannote:
Follow-up in afc5305e8: allocation exceptions are no longer silent. The renderer logs one warning with the rejected dimensions and original CanvasKit error, disables repeated retained-cache attempts, and continues through direct scene rendering. Replacing the renderer surface resets the capability so a later resize/context recovery can try again.

PR #497, opened 2026-08-12
fix(tauri): prevent Windows font loading crashes

### LINKED ISSUE #486: Renderer crashes on non-Latin text/font enumeration — Arabic, Korean, Russian all crash; list_available_fonts alone crashes too (Windows, v0.14.0)
### What happened?

Any Text node whose content needs a glyph outside the default font hard-crashes the renderer — the WebSocket automation bridge reports "Browser disconnected", often preceded by "RPC timeout (20s)", and the entire unsaved in-memory document is lost.

### Reproduction (controlled, isolated one variable at a time)

- Arabic (`معرض سيارات طرابلس`) — crashes, 3/3
- Korean (`현대 소나타`) — crashes
- Russian (`Автосалон Триполи`) — crashes
- Identical Latin text — renders fine
- Latin text with `textAlign="right"` — renders fine (rules out alignment/bidi as the trigger)
- Larger frames with more children — render fine (rules out size/complexity)
- **`list_available_fonts` (a read-only MCP tool that renders nothing) also crashes the renderer on its own.** This is the important data point — it rules out text layout/shaping entirely and localizes the fault to font enumeration/fallback resolution itself, not rendering a specific glyph.

### Platform / version

- OpenPencil desktop v0.14.0, Windows 11
- Machine has the standard Windows Arabic and CJK font sets installed
- Reproduced via MCP tool calls (`render` / `list_available_fonts`)

### Where I've traced it to (hypothesis, not confirmed — flagging for whoever picks this up)

`list_available_fonts` → `figma.listAvailableFontsAsync()` ([`src/app/automation/bridge/figma-factory.ts#L25-L35`](https://github.com/open-pencil/open-pencil/blob/v0.14.0/src/app/automation/bridge/figma-factory.ts#L25-L35)) → `listFonts()`/`listFamilies()` ([`src/app/editor/fonts/index.ts`](https://github.com/open-pencil/open-pencil/blob/v0.14.0/src/app/editor/fonts/index.ts)) → on Tauri, `invoke('list_system_fonts')` → native Rust enumeration in [`desktop/src/fonts.rs#L13-L58`](https://github.com/open-pencil/open-pencil/blob/v0.14.0/desktop/src/fonts.rs#L13-L58), which walks every installed system font via `font_kit::source::SystemSource` and calls `.load()` (DirectWrite-backed on Windows) on each one to read its weight/style.

Since Tauri runs the WebView and the Rust backend in one OS process, a native-level fault in that font-loading call (as opposed to a catchable Rust panic, which `spawn_blocking` already isolates via `JoinError`) would take the whole renderer down with it — which fits "Browser disconnected" better than a JS/WASM-layer bug. That would also explain why a read-only, render-nothing tool call can trigger it. I want to flag this as a working theory rather than a confirmed root cause — I don't have a Rust backtrace or WebView2 crash dump from the actual crash to confirm it, and I'm happy to gather one (`RUST_BACKTRACE=1`, Event Viewer, or a WebView2 crash dump) if that would help whoever picks this up trace the exact frame.

I checked for existing reports first — related but distinct: #365 (Arabic/Persian support, no repro/crash mentioned, web platform) and #324 (CJK glyph/fallback rendering issues, closed, about wrong glyphs not crashes). Neither covers an actual crash or the `list_available_fonts` angle, so filing separately; happy to have this merged into one of those threads if you'd rather track it there.

### Impact

Combined with the lack of crash recovery (#487), this makes the MCP surface unusable for any RTL or CJK design work — every crash costs all unsaved work with no warning.

dannote:
Fixes #486.

## What changed
- Return system font files through Tauri’s raw binary IPC response instead of serializing every byte as a JSON number
- On desktop, identify CJK/Arabic fallback scripts directly from text before attempting JS glyph-coverage parsing of large native fonts
- Add regression coverage for binary IPC and unknown native-font coverage
- Repair the current quality-gate regressions on `master`: use the shared `Vector` primitive, remove unsupported `SLICE` rescaling, and drop an unused test binding

## Windows reproduction and validation
Reproduced on the existing Windows Server 2025 libvirt VM: loading Korean text (`현대 소나타`) with Arial caused the WebView2 renderer to terminate with `Error code: Out of Memory`; Russian and Arabic text rendered, and `list_system_fonts` itself remained stable. A Malgun Gothic font file is ~13.5 MB, which made the old JSON byte-array IPC path especially costly.

Validated the native change on that VM: `load_system_font("Malgun Gothic")` now resolves as an `ArrayBuffer` of 13,459,196 bytes rather than a JS number array.

## Checks
- `bun run check`
- `bun run check:vue`
- `bun test tests/engine/app/font-loading.test.ts tests/engine/text/coverage.test.ts tests/engine/text/fonts/coverage.test.ts tests/engine/tauri/fonts.test.ts`
- `bun test tests/engine/figma/api/layout/plugin-compat.test.ts tests/engine/layout/auto-layout/imported-derived-layout/effective-generated-text.test.ts`
- `cargo fmt --manifest-path desktop/Cargo.toml --check`
- `cargo check --manifest-path desktop/Cargo.toml`
- `git diff --check`

The remaining two Oxlint `max-lines` diagnostics are warnings and do not fail the quality gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Windows desktop crashes when loading large system fonts for non-Latin text.
  * Improved fallback font handling so required scripts are identified before processing large font files.
  * Improved reliability when loading and registering system fonts.

* **Documentation**
  * Added an Unreleased changelog entry describing the Windows font-loading fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixed Windows desktop crashes when loading large system fonts for non-Latin text by using Tauri's binary IPC path and resolving native script fallbacks without parsing full font files in JavaScript.
+
 ### Fixed
 
 - Preserve open vector segments when the same vector network also contains filled regions. (#450)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## Unreleased
 
-- Fixed Windows desktop crashes when loading large system fonts for non-Latin text by using Tauri's binary IPC path and resolving native script fallbacks without parsing full font files in JavaScript.
-
 ### Fixed
+
+- Fixed Windows desktop crashes when loading large system fonts for non-Latin text by using Tauri's binary IPC path and resolving native script fallbacks without parsing full font files in JavaScript.
 
 - Preserve open vector segments when the same vector network also contains filled regions. (#450)
 - Match Figma Plugin API behavior for `rescale()`, page `backgrounds`, and nullable visual `absoluteRenderBounds`. (#442)
--- a/desktop/src/fonts.rs
+++ b/desktop/src/fonts.rs
@@ -112,8 +112,13 @@ fn load_system_font_blocking(family: String, style: String) -> Result<Vec<u8>, S
 }
 
 #[tauri::command]
-pub async fn load_system_font(family: String, style: String) -> Result<Vec<u8>, String> {
-    tauri::async_runtime::spawn_blocking(move || load_system_font_blocking(family, style))
-        .await
-        .map_err(|e| format!("Font load task failed: {e}"))?
+pub async fn load_system_font(
+    family: String,
+    style: String,
+) -> Result<tauri::ipc::Response, String> {
+    let data =
+        tauri::async_runtime::spawn_blocking(move || load_system_font_blocking(family, style))
+            .await
+            .map_err(|e| format!("Font load task failed: {e}"))??;
+    Ok(tauri::ipc::Response::new(data))
 }
--- a/packages/core/src/figma-api/render-bounds.ts
+++ b/packages/core/src/figma-api/render-bounds.ts
@@ -63,11 +63,7 @@ function hasRenderableGeometry(node: SceneNode): boolean {
   return node.fillGeometry.length > 0 || node.strokeGeometry.length > 0
 }
 
-function translatedBounds(
-  bounds: VisualBounds,
-  offset: Vector,
-  overflow: number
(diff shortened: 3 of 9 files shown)


dannote:
Addressed the review findings in `07f80cefd`:

- Moved the release note under `### Fixed`
- Added `MissingGraphFontScriptsOptions`
- Used the shared `IS_TAURI` constant for fallback detection

Also rebased onto current `master`; `bun run check` now passes locally.

PR #499, opened 2026-08-12
feat(app): recover unsaved documents

### LINKED ISSUE #487: New documents created via new_document (no path) get zero autosave/crash-recovery protection
### What happened?

Building a document from scratch via MCP tool calls (`new_document` with no `path`, then a long batch of `create_*`/`set_*` calls) and then losing the renderer to a crash destroys the entire document with zero recovery — no autosave, no crash-recovery journal, no "restore unsaved work?" prompt on next launch. One occurrence lost ~96 design variables and 3 component frames instantly.

For an agent-driven workflow this is worse than it sounds: a long batch of MCP calls can build a lot of state, and every crash removes all of it with no warning that it was ever at risk.

### Correction to my own assumption — autosave isn't actually absent

I initially assumed autosave didn't exist in this codebase at all. It does ([`src/app/document/autosave/create.ts`](https://github.com/open-pencil/open-pencil/blob/v0.14.0/src/app/document/autosave/create.ts), with its own E2E coverage and an open tuning issue, #306). The real gap is narrower and more specific:

`createAutosave()`'s debounced watcher no-ops whenever `hasWritableSource()` is false:

```ts
if (!hasWritableSource()) return
```

`hasWritableSource` is wired as `() => !!getFileHandle() || !!getFilePath() || !!getStorageBinding()` ([`src/app/document/io/source.ts#L72-L79`](https://github.com/open-pencil/open-pencil/blob/v0.14.0/src/app/document/io/source.ts#L72-L79)).

`handleNewDocument()` — the MCP `new_document` handler — when called without a `path` (the common case for pure in-memory/agent-driven work) just does `createTab()` and returns: no file path, no file handle, no storage binding, ever, until the user does an explicit Save/Save As ([`src/app/automation/bridge/file-handlers.ts#L32-L44`](https://github.com/open-pencil/open-pencil/blob/v0.14.0/src/app/automation/bridge/file-handlers.ts#L32-L44)).

So: **any document created via `new_document` without a path gets zero autosave protection for its entire lifetime**, silently, with nothing in the MCP tool response or the UI indicating that this document isn't being protected. That's the precise, fixable bug — not "autosave is missing."

### Ask

Two related things, happy to have them tracked as one issue or split if you'd rather:

1. **Close the gap above** — give documents without a writable source *some* form of periodic local safety net (independent of the user ever choosing a save location), so autosave protection doesn't depend on having already saved once.
2. **Graceful recovery**, not just autosave — even with (1), a crash mid-session should offer a way back. A "restore unsaved work?" prompt on next launch (checking for an orphaned recovery snapshot with no corresponding open tab) would turn "silently lost forever" into "recoverable," which matters a lot more for agent-driven sessions that can rack up a lot of state between explicit saves than it does for a human clicking around.

I can send a PR once the shape of (1)/(2) is agreed — in particular where a pre-save recovery snapshot should live (I'm leaning toward a small dedicated local store separate from the existing cloud-sync local cache in `src/app/storage/local-store/`, since that one's tied to `StorageProviderID`/sync semantics that don't apply here) and what the restore UX should look like.

Related: #486 — that crash is the other concrete way this gap costs real work.

dannote:
## Summary
- Persist debounced `.fig` recovery snapshots for source-less documents, including pathless MCP documents
- Offer restore/discard actions for orphaned snapshots when the editor starts
- Remove recovery data after successful saves or explicit tab closure
- Preserve the scene version associated with exported bytes during asynchronous saves
- Serialize recovery writes, deletion, adoption, and IndexedDB fallback to avoid stale snapshots and lifecycle races

## Validation
- `bun run check`
- `bun test tests/engine/app/document/recovery tests/engine/app/storage tests/engine/tauri/document-io.test.ts`

Fixes #487


dannote:
@coderabbitai review

dannote:
Rebased onto current `master` and addressed the remaining recovery/lifecycle findings in `5247db283`:

- Awaited all `idb` object-store requests so transaction failures cannot be silently missed
- Serialized recovery backend operations and migrate existing snapshots before switching to memory
- Kept persistent deletion failures rejected rather than treating an empty memory fallback as successful cleanup
- Made snapshot adoption wait for active writes and remove the previous recovery ID
- Made tab closure await recovery deletion before disposal/removal
- Removed duplicate store disposal
- Added deferred-write adoption coverage

Validation now passes locally: `bun run check` and 37 targeted recovery/storage/Tauri tests.

PR #524, opened 2026-08-14
fix(editor): place nested component instances in world space

dannote:
## Summary

- derive default editor instance placement from the source component's transformed world bounds
- convert the target point into the requested destination parent's coordinate space
- preserve explicit coordinates supplied by Assets and canvas drag-and-drop callers

## Behavior

Creating an instance from the editor command places it 40px to the right of the source component's world-space bounds, even when the component is nested or transformed. A transformed destination parent receives the equivalent local coordinates, and undo/redo preserves the corrected placement.

This does not change `ComponentNode.createInstance()` in the Figma Plugin API. That path continues matching Figma by creating on the current page at its default origin.

## Validation

- `bun run check`
- `bun test tests/engine/editor/undo/create-instance.test.ts`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected automatic placement of newly created instances beside nested source components.
  * Improved placement when source or destination containers have transformations.
  * Preserved explicitly chosen drag-and-drop coordinates.
  * Ensured undo and redo retain the corrected placement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 
 ### Fixed
 
+- Place editor-created instances beside nested source components in world space, including transformed source and destination parents.
 - Harden collaboration node synchronization against malformed remote source metadata and geometry while excluding derived text-renderer caches.
 - Transfer native `.fig` exports over binary Tauri IPC instead of JSON byte arrays, preventing large desktop saves from being truncated or exhausting WebView memory. (#484)
 - Keep unsaved source-less documents recoverable after their editor tab is closed, matching Figma's retained offline-change behavior.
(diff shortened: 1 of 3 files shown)


dannote:
Local browser smoke completed with `agent-browser` against the PR branch:

- Created a component nested in a translated frame.
- Invoked **Create instance** through the actual layer-tree context menu.
- Confirmed the page-level instance landed at the expected world-space position (`445, 405`).
- Exercised keyboard undo/redo and confirmed deletion, prior-selection restoration, recreated placement, and instance reselection.
- Verified a rotated destination parent receives local coordinates whose world AABB starts at the expected target point.
- Verified explicit local coordinates remain unchanged.
- Cleared runtime logs before the final pass; no page errors or new console output occurred during create/undo/redo.

PR #546, opened 2026-08-17
fix(desktop): grant missing fs-plugin permissions for exists/read_text_file/create/open

0xemc:
## Summary

`desktop/capabilities/default.json` only grants `fs:allow-read-file`, `fs:allow-write-file`, `fs:allow-mkdir`, `fs:allow-remove`, and `fs:allow-watch`/`fs:allow-unwatch`. The app also calls `exists()`, `readTextFile()`, `create()`, and `open()` from `@tauri-apps/plugin-fs` — in Tauri v2 each fs-plugin command requires its own explicit permission identifier, and `allow-exists`, `allow-read-text-file`, `allow-create`, and `allow-open` were never added.

## Impact

The most visible symptom: `EditorView`'s automation startup (`src/app/automation/mcp/spawn.ts`) calls `discoveryFileExists()` / `readDiscoveryToken()` to detect and reuse an already-running `openpencil-mcp-http` server. With `allow-exists` and `allow-read-text-file` missing, both calls silently fail (the fs-plugin IPC rejection is caught and treated as "file not found"), so the app always falls through to spawning a *redundant* MCP server process and registers with a freshly generated auth token instead of the real one — which the already-running server then rejects. Net effect: automation never actually connects to an already-running MCP server, even though `/health` reports it as available.

Confirmed by tracing the full startup path with temporary instrumentation: `hasDiscovery` was always `false` for a discovery file that unquestionably existed and was readable (same UID, `0600 mc:staff`), and adding the missing permissions immediately fixed it — `hasDiscovery=true`, correct token read, WebSocket registration succeeds, `/health` flips to `"status":"ok"`.

## Test plan

- [x] `bun run lint` / `bunx tsgo --noEmit` clean
- [x] Built `desktop/` from source (`bun run tauri build`, ad-hoc signed for local testing) and confirmed end-to-end: automation server registration succeeds and MCP tool calls (`list_documents`, `list_pages`, `get_current_page`) work through a live MCP client, which failed consistently before this change

## Summary by CodeRabbit

* **New Features**
  * Desktop automation can now reconnect to an already-running MCP server.
  * Desktop applications can read text files and check whether files exist at supported configuration locations.
  * Existing unrestricted file-writing capability remains available.

* **Documentation**
  * Added a changelog entry describing MCP server reconnection support.


dannote:
Thanks for digging into this and testing it end to end. I pushed a small follow-up to narrow the new permissions and add the changelog entry.

Really appreciate the detailed investigation!

PR #549, opened 2026-08-17
feat(figma-api): expose instance-swap component properties

0xemc:
## Summary
- Adds `figma.exposeInstanceSwap(slots, candidates, propertyName)` to the scripting API: exposes one or more nested instances as a real Figma instance-swap slot — the enclosing component (or component set, walking past intermediate variant members to find it) gets an `INSTANCE_SWAP` component property offering the given candidates, and each slot instance is tagged with a `componentPropertyReference` so it responds to that property.
- Mirrors Figma's own "Create component property > Instance swap" — this lets a designer pick which component fills a slot (e.g. swap the icon on a button between several options) instead of the slot being baked-in static content, which is the normal expectation for a reusable design-system component.
- Exposed as the `expose_instance_swap` MCP/CLI/chat tool, following the same shape as the existing `combine_as_variants` tool (#547).

## Why
Building real reusable components sometimes means an icon/content slot needs to be swappable rather than fixed. The underlying `INSTANCE_SWAP` property type and `SceneGraph.swapInstanceComponent` primitive already existed and round-trip correctly through `.fig` import — they just weren't reachable from the scripting/tools layer that MCP, the CLI, and AI chat all share. This closes that gap using the same primitives the editor's own variant-property UI relies on.

Note: this PR is independent of #548 (a `.fig` save/reload bug affecting all app-created component properties, not just this one) — either can land first, but both together are what you need for a swap slot created this way to actually survive a save/reload cycle in practice.

## Test plan
- [x] New unit tests in `tests/engine/figma/api/create/expose-instance-swap.test.ts`: adds a property to a standalone host component and tags the slot; shares one property across slots in different variants of the same component set; rejects non-instance slots; rejects non-component candidates; rejects slots from unrelated hosts.
- [x] `bun run lint` clean (0 errors).

## Summary by CodeRabbit

* **New Features**
  * Added support for exposing nested component instances as instance-swap slots.
  * Added candidate component selection and optional custom property names.
  * Supports shared instance-swap properties across component variants.
  * Registered the new `expose_instance_swap` operation for extended tooling, replacing the previous variant-combination operation.

* **Bug Fixes**
  * Added validation for missing, invalid, duplicate, or already-configured slots and candidates.
  * Prevents use of slots from unrelated component hosts.


dannote (dismissed):
Verified the requested fixes are present and CI is green.

PR #473, opened 2026-08-07
WIP: Local AI, voice input, multimodal inspection, and more efficient project navigation

victorwads:
> [!IMPORTANT]
> This pull request is intentionally a work in progress. It centralizes incremental improvements to local AI, voice interaction, conversation persistence, multimodal inspection, and token-efficient project navigation.

## Motivation

I use an Apple Silicon M3 Max with 36 GB of RAM and run medium-sized language and vision models locally through LM Studio, llama.cpp, and other OpenAI-compatible servers.

While connecting local models to OpenPencil, I encountered limitations around authentication, multimodal tool results, project navigation, desktop behavior, and token usage. Agents may inspect thousands of design nodes as flat JSON even though design files are normally organized as documents, pages, and top-level frames. In one small-to-medium project, exploratory calls accumulated more than 500,000 input tokens and could approach 10 million tokens for simple traversal questions.

The goal is to give models better project context with substantially fewer tokens by navigating pages and frames deliberately and preferring rendered images when visual context is more useful than node-level JSON.

## Current changes

- Improve OpenAI-compatible Completions and Responses support for local servers.
- Allow compatible local endpoints that do not require API keys.
- Add text, image, tool, and audio model capabilities plus a Transcription API type.
- Persist Design, Review, Fast, Vision, and Audio model assignments.
- Queue follow-up messages while the agent is processing and retain them across initialization failures.
- Persist the current conversation, including tool calls and results, and recover interrupted runs from the latest valid prefix.
- Deliver tool-result images natively through Responses and through a compatible multimodal fallback for Chat Completions.
- Show tool names, inputs, outputs, and the rendered image sent to the model in chat activity cards.
- Teach the agent to inspect documents through pages and top-level frames, with compact page discovery, paginated hierarchical node search, and visual exports.
- Add and review Brazilian Portuguese localization.
- Improve desktop text-field shortcuts and compatible-provider credential behavior.

## Planned follow-up

- Add microphone controls and recording states to the composer.
- Implement the OpenAI-compatible audio transcription adapter.
- Support local Whisper-compatible services and editable transcripts before submission.
- Add conversation history management beyond the single restored conversation.
- Continue reducing system-prompt and tool-result token overhead.

## AI assistance

OpenAI Codex materially assisted with implementation, tests, localization review, and PR documentation.

## Validation

Passed locally:

- `bun run check` — passed, including package builds, lint, typechecks, i18n, package metadata, dependency, audit, secret, architecture, tool, and duplication checks
[...]

PR #550, opened 2026-08-17
feat(app): add unified files home and progressive .fig loading

victorwads:
### Unified files home

- Adds one Home workspace for local Recent Files and configured Storage Workspace documents.
- Shows each source in its own section, using shared grid or list layouts, previews, timestamps, refresh, settings, Open, and New controls.
- Routes **File → Open Storage Workspace** to the same Home instead of maintaining a separate workspace surface.
- Adds the native **Open Recent** menu for reopening and clearing local files.

### Tabs and entry flow

- Opens the desktop app on Home.
- Keeps one permanent Home tab: opening a file creates or activates a document tab; creating a document opens a blank tab; closing the final document returns to Home instead of closing the app.
- Removes a provisional tab when a local or storage-file open fails, avoiding blank tabs.

### Progressive `.fig` loading

- Reads lightweight Kiwi page metadata before materializing the document graph, so page names appear while the full document continues decoding.

### Recent-file thumbnails

- Generates and caches a Cover preview locally after opening a `.fig`, without modifying the source file.
- Chooses an exact `Cover` page name first (case-insensitive), then the first page whose name contains `Cover`.
- Uses a locally cached Cover preview when available; otherwise falls back to the thumbnail embedded in the `.fig`.
- Preserves the currently active page as the thumbnail embedded in normally saved documents, including Storage Workspace documents.

## Why

Large `.fig` files previously left the editor without useful navigation feedback while thousands of objects were decoded. The unified Home gives local and configured remote work one consistent entry point, while progressive loading makes large documents responsive sooner.

## Validation

- Targeted page-manifest, recent-files, thumbnail, cache, and tab tests.
- `cargo check --manifest-path desktop/Cargo.toml`
- `cargo fmt --manifest-path desktop/Cargo.toml -- --check`
- Recent Files Playwright coverage using an isolated clone dev server.

## Preview

<table>
  <thead>
    <tr>
      <th>List view</th>
      <th>Thumbnail view</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>List view</td>
      <td>Thumbnail view</td>
    </tr>
    <tr>
      <th>Empty Recent Files</th>
      <th>Storage Workspace only</th>
    </tr>
    <tr>
      <td>Empty Recent Files</td>
      <td>Storage Workspace only</td>
    </tr>
  </tbody>
</table>

## New Features
- Added a Recent Files home screen with recent files and configured storage documents.
- Added grid/list layouts, thumbnail previews, file actions, refresh controls, and localized empty states.
- Added native Open Recent menu support, including opening, removing, and clearing entries.
- Added progressive `.fig` loading so page structure appears sooner.
- Added Cover-page thumbnail generation and caching without modifying source files.

## Bug Fixes
- Improved thumbnail validation and Cover-page selection.
- Prevented duplicate Recent Files tabs and handled failed recent-file opens more reliably.

## Future

See more for next contributions in https://github.com/open-pencil/open-pencil/pull/473


victorwads:
@dannote This one i'm proud of! <3

src/app/document/io/source.ts
```diff
@@ -48,7 +48,8 @@ export function createDocumentSourceActions({
   getRenderer
 }: DocumentSourceOptions) {
   function buildFigFile() {
-    return exportFigFile(editor.graph, undefined, getRenderer() ?? undefined, state.currentPageId)
+    const renderer = getRenderer()
+    return exportFigFile(editor.graph, renderer?.ck, renderer ?? undefined)
```
dannote: This changes thumbnail generation for all saves, including S3 documents. Without `state.currentPageId`, export only looks for a `Cover` page. If none exists, it embeds the 1×1 placeholder, which the S3 workspace then uses as its preview.

Could we keep the current page as a fallback? Recent Files can still prefer a locally cached Cover without changing normal saved-document thumbnails.

src/app/tabs/index.ts
```diff
@@ -124,9 +156,82 @@ function isDOMImportFile(file: File): boolean {
 
 function reusableTabStore(): EditorStore {
   const current = activeTab.value
+  if (current?.showHome) return createTab().store
```
dannote: This creates and activates a document tab before the file has opened successfully. If download or parsing fails, a blank tab remains.

Could we create the tab after a successful open, or remove it in the failure path? This applies to both local and S3 documents opened from Home.

dannote (changes requested):
This looks awesome overall. Thank you for putting so much care into it! The Home screen feels polished, Open Recent is a great desktop addition, and the page manifest and ranged thumbnail work are both strong improvements.

My main concern is how this fits with the S3 workspace. Right now we would have two separate file browsers:

- Recent Files as a Home tab
- S3 documents in the Storage Workspace route

Since both already use `useDocumentWorkspace()`, I think they should become one **Home / Files** experience with sections such as **Recent** and **Cloud**.

The underlying document models should stay separate. Local files should keep their filesystem paths, while S3 documents should keep `{ providerId, documentId }` and continue using the existing local cache and sync queue. Only the presentation and entry point need to be shared.

I’m also unsure about modeling Home as a normal tab with `showHome: true`. It still creates and activates an `EditorStore`, even though it is not a document. A distinct home tab type, or a unified route-based Home, would avoid attaching graph, undo, recovery, and editor state to it.

I left two inline comments about preserving S3 thumbnails and cleaning up tabs after failed opens. The remaining CodeRabbit comments about progressive-load rollback, recent-menu identity, and cache cleanup also look valid.

Overall, this is excellent work. Thanks again!

victorwads:
@dannote
I've made some changes. what do you think?

dannote:
@victorwads I'll do some polishing on top of that before merging

dannote (approved):
Thanks @victorwads! I pushed the polish pass we discussed on top. Looks great now 💛

victorwads:
related: https://github.com/open-pencil/open-pencil/pull/473

PR #572, opened 2026-08-21
fix(ai): report zero cache tokens as observation unless cache_control was sent

### LINKED ISSUE #453: Debug panel reports "NO CACHING DETECTED" from missing telemetry, not from missing caching
## Summary

`src/app/ai/debug/index.ts:42` emits a mechanism claim based only on two zero counters:

```ts
totalCacheRead === 0 && totalCacheWrite === 0
  ? '⚠ NO CACHING DETECTED — system prompt + tools are re-processed every step'
  : ''
```

Zero cache tokens means the provider **reported** none. It does not mean none happened, and it certainly does not establish that the system prompt and tools were reprocessed every step — which the message states as fact.

## Why the inference doesn't hold

- `supportsAnthropicCaching` in `src/app/ai/chat/transports.ts` sends `cacheControl: ephemeral` only for `anthropic`, `anthropic-compatible` and `openrouter:anthropic/*`. That is correct: `cache_control` is an Anthropic wire format and means nothing to other providers.
- OpenAI-compatible providers that do prefix caching do it automatically server-side, with nothing for the client to enable. Whether `cached_tokens` comes back depends on the provider.
- So on any non-Anthropic provider the counters are expected to be zero regardless of whether caching occurred, and the warning fires unconditionally.

The reporting path itself is fine, confirmed by feeding the adapter a canned response with `input_tokens_details.cached_tokens: 18000`:

```
model layer  -> usage.inputTokens.cacheRead              = 18000
agent layer  -> usage.inputTokenDetails.cacheReadTokens  = 18000   (the field transports.ts reads)
```

So when a provider does report cache tokens, they are recorded correctly.

## Impact

The message is unactionable — a user cannot enable caching that the client does not control — but it is not harmless, because it reads as a diagnosis. It led me to conclude prompt caching was broken and to nearly file it as "the largest cost item in every log", on a run where the true state was simply "this provider doesn't report cache telemetry". A warning that produces a confident wrong conclusion is worse than no warning.

## Suggested direction

Report the observation rather than an inferred cause, e.g.

```
No cache tokens reported by this provider (it may not report them).
```

and optionally note when the provider *is* one of the Anthropic-family cases where `cache_control` was actually sent — there, zero counters genuinely do indicate caching isn't working.

UnknownObject777:
## Why

Fixes #453. The AI debug panel claimed `NO CACHING DETECTED — system prompt + tools are re-processed every step` whenever the per-step cache counters were zero. Zero counters only mean the provider *reported* no cache tokens — for OpenAI-compatible providers prefix caching happens server-side with nothing for the client to enable, so the counters are expected to be zero regardless. The message read as a diagnosis of broken caching and produced a confident wrong conclusion (it nearly got filed as "the largest cost item in every log").

## What users will see

Copying the AI debug log no longer claims caching is broken for providers that never report cache telemetry. The zero-cache line now states the observation:

- **Non-Anthropic providers** (OpenAI-compatible, Google, etc.): `No cache tokens reported by this provider (it may not report them)`
- **Anthropic-family providers** (anthropic, anthropic-compatible, openrouter `anthropic/*`) — where `cache_control: ephemeral` is actually sent: the stronger `⚠ NO CACHE TOKENS REPORTED — cache_control was sent, so caching may not be working` remains, because there zero counters genuinely indicate caching isn't working.

## Surface area

- [x] UI — AI chat debug log copy output (`src/app/ai/debug`, `ChatPanel.vue`)
- [ ] Keyboard shortcut / CLI / API / Extension / i18n / dependency / behavior
- Behavior of the copied debug-log text is the change itself; no runtime feature behavior changes.

## Validation

- Extracted `supportsAnthropicCaching` into `src/app/ai/caching.ts`, shared by `transports.ts` (behavior unchanged) and the debug formatter.
- `bun test tests/engine/app/ai/debug-log.test.ts` → 5 pass (2 new cases: openai → observation, anthropic → warning).
- `tsgo --noEmit`, `vue-tsc --noEmit`, `bun run lint` → clean (only pre-existing max-lines warnings unrelated to this change).

## AI assistance

Claude (Anthropic) was used to help draft the implementation and the PR description.

## Summary by CodeRabbit

* **New Features**
  * Added support for Anthropic prompt caching configuration across supported Anthropic-compatible models.
  * Chat debug logs now include provider and model context for clearer troubleshooting.
  * Token usage reports distinguish unavailable cache metrics from providers that do not report them.

* **Bug Fixes**
  * Improved cache-token reporting, including clearer informational messages and warnings for Anthropic caching scenarios.

* **Tests**
  * Expanded coverage for provider-specific cache-token reporting and reset handling.

### DIFF
--- a/src/app/ai/caching.ts
+++ b/src/app/ai/caching.ts
@@ -1,0 +1,13 @@
+import type { AIProviderID } from '@open-pencil/core/constants'
+
+export const ANTHROPIC_CACHE_CONTROL = {
+  anthropic: { cacheControl: { type: 'ephemeral' } }
+} as const
+
+export function supportsAnthropicCaching(providerID: AIProviderID, modelID: string): boolean {
+  return (
+    providerID === 'anthropic' ||
+    providerID === 'anthropic-compatible' ||
+    (providerID === 'openrouter' && modelID.startsWith('anthropic/'))
+  )
+}
--- a/src/app/ai/chat/transports.ts
+++ b/src/app/ai/chat/transports.ts
@@ -7,6 +7,7 @@ import { ref } from 'vue'
 import { ACP_AGENTS } from '@open-pencil/core/constants'
 import type { ACPAgentID, AIProviderID } from '@open-pencil/core/constants'
 
+import { ANTHROPIC_CACHE_CONTROL, supportsAnthropicCaching } from '@/app/ai/caching'
 import {
   classifyAIChatError,
   classifyAIChatFinish,
@@ -37,18 +38,6 @@ type ToolLoopTransportOptions = {
   effectiveModelID: string
   maxOutputTokens: number
   reasoningEffort: string
-}
-
-const ANTHROPIC_CACHE_CONTROL = {
-  anthropic: { cacheControl: { type: 'ephemeral' } }
-} as const
-
-function supportsAnthropicCaching(providerID: AIProviderID, modelID: string): boolean {
-  return (
-    providerID === 'anthropic' ||
-    providerID === 'anthropic-compatible' ||
-    (providerID === 'openrouter' && modelID.startsWith('anthropic/'))
-  )
 }
 
 function mergeProviderOptions(
--- a/src/app/ai/debug/index.ts
+++ b/src/app/ai/debug/index.ts
@@ -2,8 +2,10 @@ import type { UIMessage } from 'ai'
 
 import { buildDebugLog } from '@open-pencil/core/tools'
 import type { ToolDebugLog, ToolLogEntry } from '@open-pencil/core/tools'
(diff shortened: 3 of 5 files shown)


dannote:
Thank you for taking the time to work on this and for identifying the misleading cache diagnostic. You were right that zero cache counters do not prove that caching is broken.

We took the fix a step further in a new PR, #588, which supersedes this PR. Instead of adding provider-specific warning copy, it records AI SDK usage as factual local diagnostics, preserves the distinction between “reported zero” and “not reported,” and adds Usage/Diagnostics settings with bounded local history.

I’m going to close this PR in favor of #588 so the original insight is included in the broader implementation. Thank you again for catching the issue and starting the fix.

Issue #575, opened 2026-08-21
CLI: `openpencil import` fails with "Bun is not defined" on Node (npm install)


phongaster:
## Summary

`openpencil import` always fails with `Bun is not defined` when the CLI is installed from npm and run on Node. `packages/cli/bin/openpencil.js` is `#!/usr/bin/env node`, but `packages/cli/src/commands/import.ts` calls `Bun.file` / `Bun.write`. Every other command works fine on Node — `import` is the only one affected.

## Reproduce

```sh
npm install -g @open-pencil/cli   # 0.14.0
printf '<!doctype html><html><body><div style="width:100px;height:100px"></div></body></html>' > seed.html
openpencil import seed.html -o seed.fig
```

```
 ERROR  Bun is not defined

    at readTextFile (/opt/homebrew/lib/node_modules/@open-pencil/cli/dist/index.mjs:1382:2)
    at importHTML (/opt/homebrew/lib/node_modules/@open-pencil/cli/dist/index.mjs:1401:21)
    at Object.run (/opt/homebrew/lib/node_modules/@open-pencil/cli/dist/index.mjs:1493:37)
```

No `seed.fig` is produced. Exit code 1.

## Cause

Three call sites, all in `packages/cli/src/commands/import.ts`:

- L39 &nbsp;`return Bun.file(requireFile(path)).text()`
- L94 &nbsp;`await Bun.write(output, `${JSON.stringify(document, null, 2)}\n`)`
- L99 &nbsp;`await Bun.write(output, result.data as Uint8Array)`

These are the only `Bun.*` references in the built CLI (`grep -c 'Bun\.' dist/index.mjs` → 3), which is why no other command is affected.

The published `dist/index.mjs` also carries `#!/usr/bin/env bun`, but it is loaded via `bin/openpencil.js` (`#!/usr/bin/env node`), so that shebang is inert and the Bun globals are never available.

## Suggested fix

Sibling commands already use `node:fs/promises` — `export.ts` L1 imports `{ mkdir, writeFile }`, `convert.ts` L1 imports `{ writeFile }`. Matching that convention in `import.ts` makes it a three-line change with no new dependency:

```ts
import { readFile, writeFile } from 'node:fs/promises'

async function readTextFile(path: string): Promise<string> {
  return readFile(requireFile(path), 'utf8')
}

// L94
await writeFile(output, `${JSON.stringify(document, null, 2)}\n`)
// L99
await writeFile(output, result.data as Uint8Array)
```

[...]

PR #576, opened 2026-08-21
fix(cli): use node:fs in import so the npm CLI runs on Node

phongaster:
### Summary

`openpencil import` always fails with `Bun is not defined` when the CLI is installed from npm and run on Node, as documented in #575. `packages/cli/bin/openpencil.js` uses a Node shebang, but `packages/cli/src/commands/import.ts` calls `Bun.file` and `Bun.write`. The command exits 1 and writes no output file. Every other CLI command works on Node — `import` was the only one reaching for Bun globals.

The fix switches those three call sites to `node:fs/promises`, which is the convention the sibling commands already follow (`export.ts` L1 imports `{ mkdir, writeFile }`, `convert.ts` L1 imports `{ writeFile }`). No new dependency, no behavior change under Bun — the command simply stops depending on which runtime starts it.

### What changed

- `packages/cli/src/commands/import.ts` — replace `Bun.file(...).text()` with `readFile(..., 'utf8')` and both `Bun.write(...)` calls with `writeFile(...)`.
- `tests/engine/cli/node-compat.test.ts` — new regression test asserting no `Bun.*` global appears in `packages/cli/src`. The existing CLI tests cannot catch this: `tests/helpers/cli.ts` spawns `Bun.spawn(['bun', CLI, ...])`, so they only ever exercise the Bun runtime.
- `CHANGELOG.md` — entry under Unreleased → Fixed.

### AI assistance

Models: Claude Opus 5

### Validation

- [x] `bun run check` — not run in full; it includes `bun audit`, `steiger`, `knip`, and Playwright stages that need a complete local environment. Ran the relevant parts instead:
  - `bunx oxlint -c oxlint.json --type-aware --type-check packages/cli/src/` → 0 warnings, 0 errors
  - `bunx oxlint -c oxlint.json tests/engine/cli/ packages/cli/src/` → 0 warnings, 0 errors
  - `bunx tsgo --noEmit` → exit 0
  - `bunx oxfmt --write .oxfmtrc.json` on both touched files → no changes
  - `bun run test:dupes` → 0 clones
- [x] Tests added — `tests/engine/cli/node-compat.test.ts`. Verified as a real regression test, not a tautology: with the fix reverted it fails and names all three sites (`commands/import.ts:39`, `:94`, `:99`); with the fix applied it passes.
- [x] `CHANGELOG.md` updated.

End-to-end check against `bun run build:packages` output, which is what npm users get:

```sh
# before this change
$ node packages/cli/bin/openpencil.js import seed.html -o seed.fig
 ERROR  Bun is not defined

# after
$ node packages/cli/bin/openpencil.js import seed.html -o seed.fig --json
{ "input": "seed.html", "output": "/tmp/seed.fig", "format": "fig", "pages": 1, "rootElements": 1 }
$ node packages/cli/bin/openpencil.js import seed.html -o seed.json -f json --json   # json path too
$ bun  packages/cli/bin/openpencil.js import seed.html -o seed.fig --json            # no Bun regression
```

Both runtimes produce documents that `openpencil info` reports identically (`1 pages, 1 nodes / 1 FRAME / Fonts: Inter`). They are not byte-identical, but neither are two consecutive runs on the *same* runtime — the `.fig` writer is nondeterministic run-to-run independent of this change.

`grep -c 'Bun\.' packages/cli/dist/index.mjs` → 0 after rebuild.

### Note on `tests/engine/cli/`

19 tests in `tests/engine/cli/` (`eval`, `analyze overlaps`) fail in my environment. These are pre-existing and unrelated to `import`: on a clean checkout of `master` with no changes the suite is 25 pass / 19 fail, and with this branch it is 26 pass / 19 fail — the same 19, plus the new test passing.

### Also noted, not changed here

`packages/cli/dist/index.mjs` is emitted with `#!/usr/bin/env bun`. It is currently inert, since the package's `bin` entry is `bin/openpencil.js` and loads it as a module. I left it alone to keep this PR to the actual defect, but it may be worth aligning if the intent is for `dist/index.mjs` to be directly executable.

## Summary by CodeRabbit

- **New Features**
  - Added Usage and Diagnostics panels with local event history, retention controls, export, and clearing.
  - Added browser clipboard support with fallback handling for copy, cut, and paste.
  - Added reusable buttons and confirmation dialogs for a more consistent interface.
  - Added localized Usage and Diagnostics content across multiple languages.

- **Bug Fixes**
  - Improved vector editing commits, cancellation, snapping, handle selection, and transformed overlays.
  - Improved tool-result status handling and error reporting.
  - `openpencil import` now runs in Node-compatible environments.

- **Documentation**
  - Updated the Unreleased changelog.


dannote (commented):
The runtime migration itself looks correct, but I would not add `tests/engine/cli/node-compat.test.ts` for this. This is a source-policy constraint, and the repository already has the `lint/plugin.js` + `open-pencil/*` custom-rule infrastructure for exactly this class of check. Please replace the regex/filesystem test with a scoped AST lint rule (for example, disallow `Bun` global/member references under `packages/cli/src/`) and enable it in `oxlint.json`. That will run on every lint invocation, report the exact offending node, and avoid a test that can be bypassed by syntax/formatting changes or that scans comments and string literals. The existing CLI integration tests can remain focused on behavior.

dannote:
The new review comments concern unrelated existing E2E tests in `create-shapes.spec.ts` and `cache.spec.ts`, outside this PR’s CLI runtime-compatibility scope. I’m leaving those for a separate cleanup. The Node `fs/promises` fix and the AST lint rule for Bun globals are the complete changes for this PR.

PR #583, opened 2026-08-22
fix(chat): treat absent MCP isError as success

imsergiobernal:
## Summary

Treat MCP tool results as errors only when `isError === true` in the OpenPencil chat UI.

## Changes

- Recognize MCP tool outputs by their `content` array.
- Render MCP results with missing `isError` as completed, matching MCP semantics.
- Render MCP results with `isError: false` as completed.
- Render MCP results with `isError: true` as errors.
- Preserve the existing error handling for non-MCP tool outputs and transport failures.

## Context

The MCP specification defines `isError` as optional and treats an omitted value as `false`. The chat UI was relying on the generic tool state and could display successful MCP calls as errors when the field was absent.

## Validation

- 32 MCP tests passed.
- `bun run lint` passed.
- `bunx tsgo --noEmit` passed.
- `bun --filter @open-pencil/mcp build` passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat message status handling for tool results.
  * Tool outputs are now more accurately identified as pending, completed, or errored.
  * Structured tool results with content arrays now display the correct status.
  * Explicit errors take priority, ensuring failed operations are clearly shown even when other result details are present.

* **Tests**
  * Added coverage for tool-result status classification, including successful and failed structured responses.

### DIFF
--- a/src/components/chat/ChatMessage.vue
+++ b/src/components/chat/ChatMessage.vue
@@ -42,7 +42,22 @@ function hasErrorOutput(part: ToolPart): boolean {
   )
 }
 
+function mcpResultState(part: ToolPart): 'done' | 'error' | undefined {
+  if (
+    part.output === null ||
+    typeof part.output !== 'object' ||
+    !('content' in part.output) ||
+    !Array.isArray(part.output.content)
+  ) {
+    return undefined
+  }
+
+  return 'isError' in part.output && part.output.isError === true ? 'error' : 'done'
+}
+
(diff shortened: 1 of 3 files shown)


dannote:
Thanks for catching this — you were right that MCP’s optional `isError` field should default to success.

I’ve applied a follow-up fix that limits the special handling to actual `mcp__...` tools and preserves the existing error handling for other tools. I also added coverage for omitted, false, and true `isError` values, plus non-MCP results with a `content` field.

PR #596, opened 2026-08-26
fix(figma-api): preserve auto-layout sizing mode when converting a node to a component

0xemc:
### Summary

After `create_component`, a frame that had HUG auto-layout sizing stops resizing when its padding is changed via `set_layout` — `spacing` still visibly works, but `padding`/`padding_horizontal`/`padding_vertical` appear to do nothing. Found while building component variants through the MCP tools; originally looked like it only showed up after `combine_as_variants`, but it reproduces from `create_component` alone.

### What changed

- Root cause: `createComponentFromNode` (`FigmaAPI`) builds the new `COMPONENT` node from a hand-picked list of fields copied off the source node. That list omitted `primaryAxisSizing`/`counterAxisSizing` (the HUG/FIXED/FILL sizing-mode fields), so they silently reset to the node-default `'FIXED'` on conversion — even when the original frame was HUG-sized.
- Once sizing is `FIXED`, later padding writes still land correctly on the raw `paddingTop/Right/Bottom/Left` fields (verified directly), but the layout engine (`packages/core/src/layout/apply.ts`) only recomputes a frame's `width`/`height` from padding when its sizing mode is `HUG`, so the box never visibly grows/shrinks. `itemSpacing` keeps working because it only affects the gap between children inside the (now-fixed) box, not the box's own size — which is why spacing "still works" while padding appears broken.
- Added `primaryAxisSizing`/`counterAxisSizing` to the copied field list.
- Added a regression test: creates a HUG-sized auto-layout frame, converts it, asserts the resulting component keeps `HUG` sizing on both axes.

### AI assistance

Models: Claude Sonnet 5 (Claude Code). Root-caused by reproducing against a live document via the MCP tools (confirmed the raw padding write succeeds but box size doesn't follow), then tracing `packages/core/src/layout/apply.ts` and `packages/core/src/figma-api/index.ts` to the missing sizing-mode copy; fix and test implemented and verified against `bun test`.

### Validation

- [x] `bun run check` — ran the individual steps this change touches directly (`lint`, `tsgo --noEmit`, `build:packages`, `test:dupes`) clean; did not run the full composite `check` script since several unrelated steps (`check:secrets`, `check:audit`) fail identically on a pristine `origin/master` checkout in this sandbox and are unrelated to this change.
- [x] Tests added: see `tests/engine/figma/api/create/component-from-node.test.ts`. Targeted file: 2/2 pass. Full `bun test tests/engine/figma`: 127/127 pass. Full `bun run test:unit`: same pre-existing failure count as pristine `origin/master` (git-LFS `.fig` fixture binaries not present in this sandbox), no new failures introduced.
- [x] CHANGELOG.md updated (Unreleased → Fixed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved auto-layout HUG sizing when converting frames into components.
  * Components now retain both primary- and counter-axis sizing settings from the original frame.


dannote:
Thanks! I added a regression test covering padding changes after conversion and fixed the formatting and merge issues on top.

PR #599, opened 2026-08-26
fix(kiwi): interpret compiled schemas instead of eval-ing generated code

valorkin:
## What changed and why

`compileSchema()` built each schema's decoder/encoder by generating JS source (`compileSchemaJS`) and running it through `new Function("exports", ...)`. That's indistinguishable from `eval` to a Content-Security-Policy: any embedder whose `script-src` doesn't include `unsafe-eval` — a sandboxed plugin host, for example — can't call `compileSchema` at all. It throws at construction time, before any schema-specific code even runs.

This replaces the eval step with a small interpreter that walks the same `Schema`/`Definition`/`Field` data `compileSchemaJS` already walks, and performs the identical read/write per field directly instead of generating and executing source for it:

- Field order, the `MESSAGE` tag-loop vs `STRUCT` sequential read, and deprecated-field skip behavior are unchanged — output is byte-identical to the previously-generated codec. This is a drop-in replacement, not a new wire format.
- `compileSchemaJS` itself is untouched and still exported, for anyone introspecting the generated source or depending on it directly.
- Only `compileSchema`'s implementation changes; its signature and return shape (`{ ByteBuffer, decode<Name>, encode<Name>, <EnumName> }`) are identical.

## Validation

- `cd packages/kiwi && bun run check` — typecheck, unit tests (including the new regression test), build, and dist smoke all pass.
- `bunx oxlint -c oxlint.json` on the two changed files — clean.
- Added a regression test (`tests/schema-runtime.test.ts`) that poisons the global `Function` constructor to simulate a CSP blocking `new Function`, then asserts `compileSchema` still compiles a schema and round-trips a message through it.

## AI assistance

Claude (Sonnet 5, via Claude Code) drafted the interpreter and test, iterating against `bun run check` and `oxlint` until both were clean. I reviewed the diff and verified it preserves the original codegen's field-by-field semantics before opening this PR.

## Summary by CodeRabbit

* **New Features**
  * Schema compilation now works in environments that prohibit dynamic code execution, including strict Content Security Policy (CSP) configurations.
  * Message encoding, decoding, validation, and round-tripping continue to work as expected.
  * Improved support for custom byte-buffer implementations.

* **Bug Fixes**
  * Invalid enum values, missing required fields, and invalid field types now produce clearer source-location errors.
  * Reserved or inherited field names are rejected safely.

* **Tests**
  * Added coverage for CSP restrictions, custom byte buffers, round-tripping, and validation error reporting.


dannote:
@coderabbitai review

dannote:
Thanks for the PR. I built on top of it by moving `compileSchema()` to a CSP-safe interpreter, preserving the existing codec behavior, removing the unused legacy code generator, and adding the required Kiwi attribution.

Issue #613, opened 2026-08-31
bug: .fig import crashes — TypeError on componentPropAssignment without value, then unbounded instance-swap repopulation cascade


wuww007:
### What happened?

`openpencil info <file>.fig` (CLI v0.14.0) crashes while importing a Figma-authored design file. Two independent bugs are hit in sequence, both in the instance-override application pass (`@open-pencil/fig`, `instance-overrides.js`):

**Bug 1 — TypeError on a `componentPropAssignment` without a `value` field**

`isEmptyPropValue` dereferences `v.boolValue` without a null check. An assignment that carries only `defID` (no `value`, and no resolvable `varValue`) makes `assignment.value` `undefined` and the whole import crashes immediately:

```
 ERROR  Cannot read properties of undefined (reading 'boolValue')

    at isEmptyPropValue (.../@open-pencil/fig/dist/instance-overrides.js:1150:11)
    at resolveAssignmentValue (.../instance-overrides.js:1153:7)
    at assignmentsToValueMap (.../instance-overrides.js:1166:23)
    at applyInstanceDirectAssignments (.../instance-overrides.js:1297:37)
    at applyComponentProperties (.../instance-overrides.js:1350:2)
    at populateAndApplyOverrides (.../instance-overrides.js:2089:23)
    at importNodeChanges (.../@open-pencil/core/dist/kiwi/fig/import.js:344:3)
    at parseFigFileSync (.../@open-pencil/core/dist/io/formats/fig/read.js:9:16)
```

**Bug 2 — unbounded instance-swap repopulation cascade (stack overflow / OOM)**

With Bug 1 guarded locally, import proceeds but never terminates: `applySwapProp` swaps a slot, `repopulateInstance` repopulates the swapped-in component's children, and `applyPropAssignments` then descends into that freshly populated subtree, where another nested instance matches the same prop-ref chain and gets swapped again. Each cycle creates *new* node ids, so it is not a ping-pong on one node but unbounded generation:

```
 ERROR  Maximum call stack size exceeded

    at repopulateInstance (.../instance-overrides.js:625:28)
    at applyOverridePatch (.../instance-overrides.js:714:3)
    at applyPatchAndMark (.../instance-overrides.js:1174:6)
    at applySwapProp (.../instance-overrides.js:1207:2)
    at applyComponentPropRef (.../instance-overrides.js:1222:4)
    at applyChildPropRefs (.../instance-overrides.js:1264:12)
[...]

PR #615, opened 2026-08-31
feat(canvas): render arrow stroke caps on lines and open vectors

Ed-Key:
### Summary

`ARROW_LINES` and `ARROW_EQUILATERAL` stroke caps are accepted by the scene-graph schema and round-trip from `.fig` files, but the renderer maps every cap through `getStrokeCapEntity`, which only knows `ROUND` and `SQUARE` and falls back to `Butt`. Imported Figma arrows lose their heads, and there is no way to draw an arrow in OpenPencil. This PR makes both arrow caps render, keeps them inside visual bounds and render margins, round-trips per-vertex caps through the `.fig` style override table, and exposes both caps in the stroke inspector.

### What changed

- New pure geometry module `packages/scene-graph/src/arrow-caps.ts` (exported as `@open-pencil/scene-graph/arrow-caps`): collects open path endpoints (vertices used by exactly one segment) with per-vertex `strokeCap` overrides falling back to the stroke cap, derives the outward angle from the terminal bezier tangent, and computes weight-scaled head geometry. Head side length is `4 x` stroke weight (equilateral triangle for `ARROW_EQUILATERAL`, two 30-degree wings of the same reach for `ARROW_LINES`), chosen to sit visually close to Figma's rendering.
- `drawArrowHeads` in `packages/core/src/canvas/strokes.ts` draws heads in the stroke color: filled triangle via `fillPaint`, stroked wings via `strokePaint` with the dash effect cleared, so dashed shafts keep solid heads. Heads draw in one place, after the shaft, per stroke in `paintNodeStrokes`, so they survive every shaft path: plain and dashed centerlines, filled open vectors, non-center stroke alignment, and corner radii.
- Arrow heads reach up to `4.5 x` weight from a path endpoint (head side plus wing stroke), far wider than the `weight / 2` pad centered strokes get, so `strokeOverflow` and a new `arrowCapOverflow` account for arrow caps, including vertex-only caps, and every bounds consumer passes the node context: visual bounds, `absoluteRenderBounds`, the overlap analyzer, and the picture/raster margin in `renderShape`. Without this, heads get cropped by cached pictures, raster exports, and culling.
- Per-vertex stroke caps now survive the `.fig` vector network blob: `StyleOverride` carries `strokeCap`, `buildStyleOverrideTable` keys overrides by mirroring and cap together, and decode restores the vertex cap. Import no longer promotes a vertex cap to the node-wide cap, so a one-ended Figma arrow stays one-ended instead of growing a second head.
- The stroke inspector cap picker gains both options ("Line arrow", "Triangle arrow") with translations for all eight locales, and the app-side cap handler accepts the full `StrokeCap` union through a shared `isStrokeCapValue` guard (it previously filtered the new values out, which would have made the picker options dead). Applying a node-wide cap also clears per-vertex overrides, so the picker can always change an imported one-ended arrow instead of being silently overridden by its vertex cap.
- Playwright canvas snapshot `arrow-stroke-caps` (its own spec, `tests/e2e/canvas/arrow-caps-visual.spec.ts`) covering LINE at two weights, both cap kinds, dashed shafts, rotation, an open polyline with a node-level cap, a curved one-ended per-vertex arrow, a filled open vector, an outside-aligned stroke, and a closed path as the no-heads control.

Deliberate scope boundaries, happy to follow up on any of them:

- Heads overlay the shaft rather than trimming it. Invisible for opaque strokes since the head is wider than the shaft; a translucent stroke shows slight darkening inside the head.
- SVG/PPTX exporters still export the bare shaft.
- A LINE node applies its single `strokeCap` to both ends; per-end arrows work through per-vertex caps on VECTOR networks.

### AI assistance

Models: Claude Fable 5 (implementation and tests), GPT-5.6 via Codex CLI (independent pre-PR review rounds; its findings drove the visual-bounds accounting, the per-vertex cap round-trip and import asymmetry fix, the app-side cap guard, and the snapshot coverage).

### Validation

- [x] `bun run check` (gitleaks installed for `check:secrets`: no leaks found)
- [x] Tests added or updated: `tests/engine/geometry/arrow-caps.test.ts` (endpoint collection, per-vertex overrides, closed-loop exclusion, bezier tangents, head math, overflow accounting including vertex-only caps), `tests/engine/render/canvas/arrow-heads-draw.test.ts` (paint wiring), `tests/engine/vector/blob/roundtrip.test.ts` and `packages/fig/tests/node-change.test.ts` (vertex cap round-trip and one-ended import), `tests/engine/vue/controls/stroke.test.ts` (cap value guard), and the `arrow-stroke-caps` Playwright snapshot (generated on darwin, verified stable with a compare run). Full engine suite and all package suites run with LFS fixtures pulled; the only failures are 12 that reproduce identically on a clean checkout (component-property GUID round-trip, matching #613, plus environment-bound collab/MCP-port/font tests).
- [x] CHANGELOG.md updated

https://claude.ai/code/session_012zjEbyLSBYyPpJP4fu1XNB

## Summary by CodeRabbit

* **New Features**
  * Added line and triangle arrow stroke caps for lines and open vector paths.
  * Added arrow cap options to the stroke cap picker with localized labels.
  * Preserved per-vertex arrow cap settings in vector networks.
* **Bug Fixes**
  * Improved bounds and rendering so arrowheads remain fully visible on dashed, curved, rotated, and outside-aligned paths.
* **Tests**
  * Added coverage for arrow geometry, rendering, bounds, serialization, and stroke controls.


dannote:
Thanks for the implementation! I prepared a small follow-up commit with a few polish fixes:

- Include arrowheads in descendant visual bounds.
- Use distinct icons for the three relevant cap types.
- Avoid drawing arrowheads for effectively zero-length lines.

Issue #610, opened 2026-08-30
fig: restore canonical Figma import/export fidelity


dannote:
## Problem

OpenPencil's canonical Figma import/export pipeline currently introduces semantic drift during plain `.fig` round trips. Issue #578 reports 3,303 changed nodes in `material3.fig` without user edits.

The affected pipeline is:

```text
Figma NodeChange/Kiwi
  → decode
  → @open-pencil/fig conversion
  → SceneGraph
  → @open-pencil/fig export conversion
  → encode
```

The goal is to make this canonical pipeline faithful for the supported Figma model.

## Goal

Reduce semantic drift by tracing each affected field end-to-end:

- Kiwi decoding;
- NodeChange conversion;
- SceneGraph representation;
- synchronization and override application;
- NodeChange export;
- GUID/reference remapping;
- re-import stability.

The first targets are the drift categories documented in #578:

- instance descendant override resynchronization;
- text layout and typography;
- visibility;
- blend modes;
- vector geometry;
- structural and component references.

## Implementation principles

- Model supported Figma fields semantically and type-safely.
- Fix decode/import/export behavior rather than masking drift with source passthrough.
- Keep SceneGraph format-neutral.
- Preserve opaque source metadata only as a bounded fallback for fields not yet modeled.
- Do not treat unchanged-byte passthrough as proof of edited-save fidelity.
- Every field fix must include import, export, and round-trip coverage.
- Use semantic comparisons for remapped GUIDs and IDs.
- Validate representative files in Figma where possible.

## Acceptance criteria

- Plain round trips of representative fixtures have no unexplained semantic drift.
- Each remaining intentional difference is classified and tested.
- Repeated round trips are stable.
- Instance/component references survive import/export.
- Text layout, visibility, blend, vector, and structural fields remain faithful.
- Unsupported fields do not overwrite modeled values or silently corrupt documents.
- Browser, Tauri, CLI, and headless paths use the same conversion behavior.

## Related work

- #578 — measured `material3.fig` round-trip drift.
- #551 — instance-descendant override persistence.
- #605 — semantic INSTANCE_SWAP comparison.
- #611 — Figma component-property API compatibility.
- #495 — SLOT component properties.

The source-backed archive/journal/patch proposal previously described here is not the primary solution. It may be revisited as a bounded fallback for fields that remain intentionally unsupported after the canonical model is improved.

Issue #578, opened 2026-08-22
fig: 3,303 nodes drift on a plain read/write round-trip of tests/fixtures/material3.fig


phongaster:
## Summary

A plain `openpencil convert material3.fig -o out.fig` — no editing, no app, read then write — changes **3,303 of 158,478 nodes** in the in-repo `tests/fixtures/material3.fig`. No nodes are added or lost and every variable survives intact, but geometry, text layout, and visibility drift on a small fraction of nodes, including six buttons that become `visible: false`.

The smaller `tests/fixtures/gold-preview.fig` (38,325 nodes, 320 variables) round-trips with **zero** differences, so this only shows up at design-system scale.

## Reproduce

Both fixtures are already in the repo (Git LFS).

```sh
bun run build:packages
git lfs pull --include="tests/fixtures/*.fig"

openpencil convert tests/fixtures/material3.fig -o /tmp/m3.rt.fig
```

Then compare the two documents. I walked both with `openpencil eval`, capturing ~60 properties per node keyed by tree path, normalising numbers to 4 decimals and stripping node ids (ids are legitimately renumbered on write).

## What changes

```
2064x  y: 1 -> 0.5                                LINE nodes ("Divider")
 747x  textAutoResize: NONE -> WIDTH_AND_HEIGHT
 268x  blendMode: NORMAL -> PASS_THROUGH
  89x  textAlignHorizontal: CENTER -> LEFT
  86x  textAlignVertical: CENTER -> TOP
  64x  fontSize changed (20->14, 14->24, 24->14, 14->32)
  44x  vectorPaths changed
  19x  width   ·  15x visible (true -> false)  ·  15x lineHeight (null -> 66.7)
  14x  x       ·  13x height  ·  7x layoutGrow  ·  4x fills  ·  3x primaryAxisSizingMode
```

Concrete examples, by path:

```
page12/6:Button groups/16:Standard button group/4:Button 05
    visible: true -> false                       (and Button 06, Button 07, in two groups)

page12/6:Button groups/16:Standard button group/4:Button 05/0:Content/0:State-layer/0:Icon/0:icon
    fontSize: 20 -> 14

page0/290:Navigation bar/0:Configuration=icon & label, Segments=3/2:Segment 3/0:icon-container/0:state-layer/0:Icon/0:icon
    textAlignHorizontal: CENTER -> LEFT
    textAlignVertical:   CENTER -> TOP

[...]

PR #620, opened 2026-08-31
fix(fig): preserve live instance descendant overrides

dannote:
## Summary

- Preserve live edits to nested instance descendants during FIG override synchronization.
- Record descendant field overrides through the shared proxy update path.
- Keep live fill and text values ahead of stale imported symbol overrides during re-synchronization.
- Serialize nested fill overrides during FIG export.

This addresses the instance-side drift identified in #578 and advances the canonical import/export fidelity work tracked in #610.

## Validation

- `bun run format`
- `bun run build:packages`
- `bun run lint:structure`
- `bun run check:arch`
- `bun run test:dupes`
- Focused instance override tests: 4 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved instance editing so customized fills remain intact when components synchronize or are reloaded.
  - Prevented older stored overrides from overwriting newer live edits.
  - Preserved source component fills while retaining nested instance customizations.

- **Improvements**
  - Improved export and reload support for instance fill overrides.
  - Preserved instance-swap settings, preferred values, and assignments across save/reload cycles.
  - Added coverage for fill override persistence and instance-swap round trips.


dannote:
Superseded by the consolidated implementation in #551.

PR #621, opened 2026-08-31
test(fig): place stroke import coverage with current imports

dannote:
Move the stroke import coverage out of the legacy test namespace. The cases exercise current `.fig` import behavior, including Figma-compatible odd-length dash patterns, so they belong alongside the other current import tests.


dannote:
Closing this follow-up; the original change is already merged, and this test-only location cleanup is not worth carrying as a separate PR.

PR #632, opened 2026-09-01
docs: add one-click deployment option (RepoCloud and Render)

cosark:
Adds section and buttons for one-click deployment providers to the README.

## Summary by CodeRabbit

* **Documentation**
  * Added a Deployment section to the README.
  * Included one-click deployment options for RepoCloud and Render.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ See the [roadmap](https://openpencil.dev/development/roadmap) for product direct
 
 One-click deployment providers: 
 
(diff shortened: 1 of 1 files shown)


dannote:
Thanks, but we’re not ready to advertise third-party one-click deployment providers. OpenPencil Cloud and the supported self-hosting model are still being developed in #636, and deployment options need to be based on maintained, tested configuration for that architecture.

The proposed Render link also has no corresponding `render.yaml`, so it does not currently provide a working deployment. We’ll document supported deployment targets as part of the Cloud release rather than add provider badges separately.

PR #656, opened 2026-09-07
fix(ui): keep button labels on one line

itorr:
## Summary

- prevent button labels from wrapping when surrounding layouts constrain their width
- keep the default in Tailwind’s base layer so an explicit whitespace utility can override it
- cover constrained CJK labels with a browser regression test and document the fix

## Testing

- `bun run format`
- `bun run build:packages`
- system Chrome computed-style assertion (`nowrap` by default, `normal` with an explicit override)
- `bun run check` (all completed checks passed; `check:audit` was blocked by the configured npm mirror returning 404, and `test:tools` was blocked by oxlint fixture subprocesses exiting without JSON)
- `bun run test:dupes`

## Summary by CodeRabbit

* **Bug Fixes**
  * Button labels now remain on a single line when controls have limited horizontal space.

* **Tests**
  * Added coverage to verify button text does not wrap under constrained layouts.

* **Documentation**
  * Updated the unreleased changelog with this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 
 ### Fixed
 
+- Keep button labels on one line when controls are constrained by surrounding layouts.
 - Resolve `$name` references in imported `.pen` fills, stroke fills, font families, dimensions, and spacing without requiring a `--` prefix. (#563)
 - Prevent the stock photo tool from replacing text, lines, structural layers, or containers with content while supporting closed shape geometry.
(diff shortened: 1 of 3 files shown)


dannote:
Superseded by #670, now merged. AppButton owns the nowrap default while allowing explicit wrapping overrides, with a regression test for constrained CJK labels and an explicitly wrapping label. This covers the intended fix without a global rule for every button.

PR #679, opened 2026-09-11
fix: prevent unbounded duplication of Figma-imported instance children

joeycumines:
### Summary
Fixes unbounded duplication when editing Figma-imported components. Imported instance children arrive with `componentId: null`, so every component edit re-cloned them. This compounded per keystroke into runaway memory. Also fixes sibling inversion, extras jumping to front on sync, and stack overflow on self-referential components.

### What changed
- **Sync:** match unmatched children by name+type (forward/FIFO) instead of cloning; guard self/cyclic sync; keep unmapped extras at the end.
- **Import/paste:** link children positionally at import time so renamed children stay linked; clipboard paste is scoped to new instances only (existing instances untouched).

### AI assistance
Models: Muse Spark 1.2

### Validation
- [x] `bun run check` passes on changed files (pre-existing `check:arch` / `check:packages` failures only, unrelated)
- [x] Tests added: 16 new tests (13 sync + 2 paste-scope + 1 linkage), all pass; existing suites green (246 + 46 + 147)
- [x] CHANGELOG.md updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicated children when editing or synchronizing Figma-imported and pasted components.
  * Preserved extra instance children without unexpectedly reordering them.
  * Improved matching for renamed or serialized children while retaining positional overrides.
  * Prevented pasted components from relinking or changing pre-existing instances.
  * Added safeguards against recursion issues in self-referential or cyclic component structures.


dannote:
Hi Joseph, thank you for following the project so closely. This is exactly what we need for one of the few remaining release blockers. I’ll push a small fix on top for the extra same-type child case and merge it soon.

joeycumines:
No worries @dannote , I'll leave it to you 😄

Issue #674, opened 2026-09-10
bug: list_available_fonts tool causes OOMs the WebView2 renderer


LZL0:
### What happened?

`enumerate_system_fonts()` calls `f.load()` on every face of every installed family — just to read two properties, `props.weight` and `props.style`:
```
.filter_map(|f| {
      f.load().ok().map(|font| {
          let props = font.properties();
          // ...only reads weight + style
```
On Windows `f.load()` is DirectWrite-backed and pulls the entire font file into memory. You have 849 font files, and CJK faces run ~13.5 MB each (PR #497's own number for Malgun Gothic). So it loads potentially hundreds of MB to extract a weight integer and an italic flag — and OOMs the WebView2 renderer before FONT_CACHE can ever populate, which is why it crashes every time rather than once.

It's the same class of bug #497 fixed — needlessly moving whole font files — just in the enumeration path the maintainer never reproduced. spawn_blocking guards against a Rust panic but not a native DirectWrite fault, which is exactly what the issue's author suspected.

### Platform

Windows Desktop

### Steps to reproduce

enumerate_system_fonts() tool use

### Affected area

Desktop app

### Version or build

latest HEAD

### Console output or logs

```shell

```

### Before submitting

- [x] I searched existing issues and did not find a duplicate.

PR #682, opened 2026-09-12
fix: load fonts with request in desktop app cause OOMs (replaced with fetch)

them7d:
### Summary

loading fonts with fetch instead of request because it cause memory leak leading webview failed as video show and add if condition for checking if buffer aborted before converting buffer to unit8array
Related to #674; the larger-font-library enumeration case remains open. 
### What changed

- load fonts with fetch instead of request
- checking buffer before converting it to unit8array
- remove `createTauriFetch` there is no use for it just in test file
### AI assistance

deepwiki used for tracing functions and understanding the flow of data

Models: deepwiki, GLM5.3

|before|after|
|----------|----------|
| https://github.com/user-attachments/assets/ad47bf82-7121-4768-813e-9fe2a3dffae7 | https://github.com/user-attachments/assets/768cc993-3666-406a-8b69-8a5fcbf7c1c7 |

### Validation

- [x] `bun run check`
- [x] Tests updated because: remove `createTauriFetch` test unit
- [x] CHANGELOG.md updated

## Bug Fixes

- Fixed recursive desktop proxy requests that could occur when font downloads intercepted internal app traffic.
- Improved routing for internal desktop requests to use the appropriate native network path.
- Improved cancellation handling during desktop network operations.

## Documentation

- Added an unreleased changelog entry documenting the desktop font-loading fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@
 
 ### Fixed
 
+- load fonts with request in desktop app cause OOMs (replaced with fetch)
+
 - Preserve edited instance text, including cleared labels, when saving and reopening `.fig` files.
 - Honor `.pen` frame layout defaults and sizing and padding shorthands so imported auto-layout frames keep their computed dimensions and child positions. (#564)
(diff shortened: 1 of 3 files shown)


dannote:
Thank you! 💛

PR #714, opened 2026-09-16
feat(settings): configure tool access, MCP failures, and step limits

### LINKED ISSUE #573: feat: customisable max agent steps
### Problem

The current max agent steps is hard-coded to 50. It may be enough for a simple design but might not to design with complex requirements.

https://github.com/open-pencil/open-pencil/blob/0f981ff2afadfb2c19f76a63fdd617ea3092c5fb/src/app/ai/tools/index.ts#L21

https://github.com/open-pencil/open-pencil/blob/0f981ff2afadfb2c19f76a63fdd617ea3092c5fb/src/app/ai/chat/system-prompt.md?plain=1#L222

### Proposed solution

Allow customisation of `MAX_AGENT_STEPS` inside the settings modal.

I am happy to implement it if this gets approved.

### Affected area

App UI

### Alternatives or workarounds

na

### References, examples, or mockups

_No response_

### Before submitting

- [x] I searched existing issues and did not find a duplicate.

### LINKED ISSUE #584: feat: add tools toggle to setting modal
### Problem

The tools availability is fixed (hardcoded), LLM able to utilise all tools in the core and only `get_components`, `list_libraries`, `insert_library_component` from `EXTENDED_TOOLS` set.

https://github.com/open-pencil/open-pencil/blob/f75d67ad4d239da7ed3ba2cd395e21fef983c1eb/src/app/ai/tools/index.ts#L100-L102

Say, a user wanna to allow the LLM to create components, it is no way to do so right now.

I am happy to implement it if this gets approved.

ps: I am not sure whether it is intended or not. Based on my observation, there are a ton of tools defined in `EXTENDED_TOOLS`, but not all of them are allowed to be used by the LLM. Please correct me if there is any misunderstanding.

### Proposed solution

Add a new tab to the settings modal with a list of tools' toggles.

### Affected area

App UI

### Alternatives or workarounds

na

### References, examples, or mockups

_No response_

### Before submitting

- [x] I searched existing issues and did not find a duplicate.

dannote:
### Summary

Built-in AI could only use a hardcoded subset of the shared tool registry, so extended tools such as `create_component` were unreachable, and the maximum agent steps was a constant that users could not raise for longer tasks (#584, #573). Sidebar settings now separate agent **permissions** from **MCP endpoints**: a **Tool access** page configures which tools built-in AI and the local MCP server may use, and an **MCP** page holds the local server, browser access, and MCP connections.

### What changed

- Tool permissions come from the shared registry plus explicit exposure and execution checks instead of a name list. Built-in AI keeps its compact shipped tool set as the default and stores per-user overrides; local MCP keeps its existing disabled-name preference. Group switches follow the catalog rather than search results, and Restore defaults restores only the active target.
- The Tool access page owns a fixed toolbar (target, enabled count, Restore defaults, search) above the dialog body, so the list uses the full height instead of a nested scroll region that showed three rows, and no row is clipped. The list became a presentational component over the existing `useToolAccess` composable.
- Each AI message captures the saved step limit and uses it for the agent stop condition, remaining-step warnings, and limit detection, so changing Settings does not alter an in-flight request. Tool requests are rebuilt per message, and disabled extended tools still validate when they appear in conversation history.

### Related fix

The local MCP server reported one generic health failure for every cause. Failures are now classified (not installed, denied command, early exit, startup timeout, rejected token, unexpected response, unreachable) and rendered from the translation catalog. The payload sits in a collapsible inside the alert — unmounted while collapsed, so the live region announces only the summary — with a copy action for issue reports. The desktop lookup also searches the common global binary directories that a graphical launch omits from `PATH`, and reports where it looked.

### AI assistance

Models: deepseek-v4.1-flash

### Validation

- `bun run check` — passed.
- `bun run test tests/e2e/settings tests/e2e/app/webmcp.spec.ts --project=openpencil` — 43 passed at 1280px and 390px, covering both pages, the dirty-form guard, and the Russian-locale feedback case.
- `bun test tests/engine/app/chat tests/engine/app/settings tests/engine/tools/ai-adapter.test.ts` — 89 passed, covering per-message tool and step-limit refresh, invalid stored values, and target independence.
- `bunx playwright test --project=storybook-chromium tests/e2e/storybook/controls.spec.ts -g settings-tool-access-list` — passed in light and dark.
- `bun test tests/engine/app/automation tests/engine/tauri tests/engine/mcp` and the settings/chat suites — 302 passed, covering failure classification, health-probe diagnostics, and per-message refresh.
- `bun run test tests/e2e/settings/mcp-failure.spec.ts` — passed: a failing health probe shows the classified English copy and the same reason in Russian, with no generic message remaining.
- Manual browser inspection: desktop and 390px Tool access and MCP pages, fixed toolbar while scrolling, and the Chat step-limit field.
- `cargo test --lib` in `desktop` — 9 passed, including the new `PATH` merge tests; `cargo clippy --all-targets` reports only a pre-existing warning.
- `bunx playwright test --project=storybook-chromium tests/e2e/storybook/controls.spec.ts -g design-system-disclosure-collapsible` — passed in light and dark.
- Not run: full `test:unit` and `test:storybook` suites, native desktop checks, and real-provider AI chat.
- Changelog: `Unreleased` gains one `Added` outcome per feature.

Closes #584
Closes #573


dannote:
@coderabbitai review

Re-review the current head. Commit 99c3df9f5 restructured the settings pages after your previous review of 31e4d14d1: the Automation page was split into separate MCP and Tool access pages, the tool list became presentational over the existing composable, and the page now owns the only scroll container.

dannote:
@coderabbitai full review

The incremental re-review was skipped because the earlier request arrived after the first commit, so commits 99c3df9f5 through b3baa64e3 were never reviewed. Please review the entire changeset.

dannote:
@coderabbitai full review

PR #717, opened 2026-09-17
feat(demo): rebuild the first page as a component library with staged loading

dannote:
### Summary

The demo built its document behind an empty canvas with no indication anything was happening, its first page still carried the legacy reference page in a style that matched neither the current showcase nor pages 02 and 03, and an interrupted build left that document half-built.

### What changed

- Demo generation reports the same preparation phases as a document open, so the existing canvas overlay and tab indicator cover it, and a superseded or failed build rolls back the pages, sections and variables it created instead of leaving a half-built document.
- The first page becomes a component library in the current design-jsx style — a Button set with two variants, linked instances, text and boolean component properties, and the variable collections. The legacy reference page and its app-preview, standalone and effects sections are removed; pages 02 and 03 already cover that ground.
- Capability notices now point at one neutral support reference through a shared external-link primitive that SettingsLink also uses, and `SafariBanner` is renamed `FileApiBanner` because it tests the File System Access API rather than Safari.

### AI assistance

Models: DeepSeek V4.1 Flash (deepseek/deepseek-v4.1-flash)

### Validation

- `bun run check` — exit 0.
- Browser: demo-workspace (including a new abandoned-build test, verified to fail with the rollback disabled), layers, colour-picker, stroke-picker, assets-panel and the P3 spec pass.
- Known: the pages 02 and 03 demo snapshots differ by 735 px locally. That is identical before this change and on pristine master — an environment difference in this checkout, not a regression from this work.
- Not run: the full Playwright suite; no workflow invokes browser E2E.
- Changelog: updated under `Unreleased`.


dannote:
Addressed in `338a858c7`. The build now tracks the pages, section roots, and variable collections it creates and removes them whenever it does not finish, restoring the original page name and clearing its variables.

I chose rollback over the two alternatives you suggested: a staging graph would require rewriting the section builders, which are store-based (`store.createShape`, `store.select`), and changing supersede semantics in the shared preparation controller would add document-replacing policy for one caller. Rollback also covers a closed tab and a failed build, and it is guarded by graph identity because node and page IDs are only unique within one graph — a document that replaced the demo is never touched. The new e2e test interrupts generation mid-build after it has already added its pages, and I verified it fails with the rollback disabled.

On the second finding — making the loader observation deterministic — I am leaving it as is. The overlay is visible for the whole multi-second generation window, and asserting transient DOM state with a locator before the interaction ends is what `AGENTS.md` prescribes; the 300 ms leave transition is not load-bearing because generation runs far longer than that.

The `PNUM` warning still visible in this branch's logs is fixed separately in #716 (`3f924aa9a`); it comes from the shared `.fig` export path.

PR #725, opened 2026-09-18
feat(MCP): follow agent activity in canvas

victorwads:
## Summary

Add an opt-out **Follow agent** setting for the local MCP server. When enabled, the OpenPencil UI follows the document tab, page, and layers that an MCP agent just inspected, selected, or exported, so people can see the context the agent is using.

## What changed

- Persist a default-on Follow agent setting in the MCP Settings panel.
- After a successful MCP Core tool, activate its target document tab, switch to its target page, and fit its requested or returned layers in the viewport. Page-level operations fit the target page instead.
- Cover selection/export focusing and page switching, and document the behavior for MCP users.

## Validation

- `bun test tests/engine/mcp/server/index.test.ts tests/engine/app/automation/follow-agent.test.ts` — passed (26 tests).
- `bunx tsgo --noEmit --pretty false` — passed.
- `bun run check:changelog` — passed.
- `bun run check:docs` — passed.
- `bunx oxfmt --check src/app/automation/mcp/follow-agent.ts src/app/automation/mcp/preferences.ts src/app/automation/bridge/tool-handlers.ts src/components/settings/mcp/MCPSettingsPanel.vue packages/vue/src/i18n/messages/automation.ts tests/engine/app/automation/follow-agent.test.ts` — passed.
- Full `bun run check`, `bun run test:unit`, and `bun run test` were not run; this is a focused MCP bridge, Settings, and viewport change.

## AI assistance

GPT-5.6-terra (DeepSeek Harness).


victorwads:
@dannote What do you think?

I had this on my other infinity branch hahah, and then i needed the improvements from master, so i'm passing the features i miss most.

dannote:
@victorwads ha, you're reading my mind. Following agent activity is one of the things I really wanted in the next major release, so the timing is great.

Longer term I want to unify how humans and agents show up in the canvas. Right now collaborators have cursors and agents have no presence at all, and I'd rather have one model for both. This is a nice reference for the smooth cursor part: https://liveblocks.io/blog/how-to-animate-multiplayer-cursors#splines

Also related, #692 is my own work on streaming generated elements, which is really the other half of this.

I did consider the openui.com approach, but it looks less token efficient and adds extra cognitive load on agents.

Thanks for all the contributions lately. If you want to coordinate on roadmap stuff, I finally opened a Discord: https://discord.gg/N9fMAa22. It's empty since I hadn't advertised it, but I think it's time now.

victorwads:
Ohh @dannote, It's me that should be thanking you, i'me an independent developer starting my on company, i do not have money to pay Figma together with other tools, soo, Open Pencil is helping me a lot!

I've entered on discord, but no messages permission

victorwads:
@dannote There are two new tools I’m planning to add to my version, since sometimes the AI just needs to see one thing instead of navigating through the whole document—for example, when it only needs to fix a margin, color, or font according to a spec:

- **`see_user_selection`** returns a PNG of the current selection for vision, plus a compact node tree up to depth 2 with the relevant IDs.
- **`get_user_selection_details`** accepts specific node IDs and returns richer structured properties with a limited depth, allowing the AI to inspect only the relevant branches progressively.

Together, they provide an image-first workflow while keeping the amount of structural context under control.

## The problem

I’m considering a restricted mode that exposes only these two tools. In this mode, the AI would have no access to any other document, navigation, or editing tools—it could only see the user’s current selection and progressively inspect specific nodes within it. This would provide a focused, predictable, and privacy-conscious workflow where the user explicitly controls the AI’s entire design context through their selection. Since both tools are read-only, the AI could analyze the selection and recommend precise changes; applying those changes would still require the user or an explicitly enabled editing tool.

User says: "Look my open-pencil/figma with me and fix the spaces ans colors as specified"

## Why

This is conceptually similar to Figma MCP’s get_screenshot + get_metadata/get_design_context flow, but the goal is to combine the initial screenshot and bounded structural context in one call, and optionally expose only these two tools so the user’s selection becomes the AI’s entire accessible design scope.

